PR #1, opened 2024-03-01
[outdated] POC H265 Support, Codec Detection, and Android Build

seanavery:
- Added `avformat` lib as dependency.
- Created wrapper around `avformat_find_stream_info` for codec detection.
- Separated out codec init and rtp handlers for h264/h266.
- Added AppImage wrappers for Linux targets.
- Added Android specific build scripts for FFmpeg and module.

## Testing
- Linux Arm64 AppImage tested on RPI4 and Wansview RTSP camera that has H264 and H265 tracks.
- Android module tested in emulator (Android 10 Q API 29).


cmd/module/cmd.go
```diff
@@ -26,7 +26,7 @@ func realMain() error {
 		return err
 	}
 
-	err = myMod.AddModelFromRegistry(ctx, camera.API, viamrtsp.ModelH264)
+	err = myMod.AddModelFromRegistry(ctx, camera.API, viamrtsp.Model)
```
erh, line 29: can we add the other as an alias or something so as not to break existing users?

meta.json
```diff
@@ -1,13 +1,13 @@
...
-  "url": "https://github.com/erh/viamrtsp",
+  "module_id": "viam-labs:viamrtsp",
+  "visibility": "private",
+  "url": "https://github.com/viam-labs/rtsp-camera",
   "description": "module for rtsp, currently supports x264",
   "models": [
     {
       "api": "rdk:component:camera",
-      "model": "erh:viamrtsp:rtsp-h264"
+      "model": "viam-labs:viamrtsp:rtsp"
```
erh, line 9: let's not move yet
we'll do that separately later.

run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ $(command -v getprop) ]; then
+	echo detected android host, reading libraries from $PWD/lib
+	export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
+fi
+
+exec ./bin/viamrtsp-android-arm64 $@
```
erh, line 8: is this always supposed to say android?

Makefile
```diff
@@ -1,6 +1,138 @@
...
+build-android:
+	# if this fails with Camera interfaces, run `make edit-android` first
+	GOOS=android GOARCH=arm64 CGO_ENABLED=$(CGO_ENABLED) \
+		CGO_CFLAGS="$(CGO_CFLAGS)" \
+		CGO_LDFLAGS="$(CGO_LDFLAGS)" \
+		CC=$(CC) \
+		go build -v -tags no_cgo \
+		-o $(OUTPUT) ./cmd/module/cmd.go
+
+module.tar.gz: $(OUTPUT) run.sh
```
erh, line 61: should the make module: at the bottom be removed now?

run.sh
seanavery: No, `run.sh` entry is only used for Android right now. I have been using the appimage builds for testing on Linux.

Working on a patch that will use run.sh for linux as well - will test out and update PR with unified entry.

cmd/module/cmd.go
seanavery: Reverted model name back to `rtsp-h264`.

run.sh
seanavery: Ok, updated so that Linux/Android module targets both work from `run.sh` entry.

cmd/module/cmd.go
erh: Now it's a lie :) 
Can't we have 2 names for the same thing for now

Makefile
```diff
@@ -1,6 +1,127 @@
...
+
+# CGO settings
+CGO_ENABLED := 1
+CGO_CFLAGS := -I$(FFMPEG_PREFIX)/include
+CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
+
+# Output settings
+OUTPUT_DIR := bin
+OUTPUT := $(OUTPUT_DIR)/viamrtsp-$(GOOS)-$(GOARCH)
+APPIMG := rtsp-module-$(MOD_VERSION)-$(ARCH).AppImage
```
erh, line 41: do we really need appimage? 
really trying to move away from it.

cmd/module/cmd.go
randhid: We can @seanavery you'd need to register:

```var Model = family.WithModel("rtsp")``` in rtsp.go now, and add a new 

`Register` function to init with the same constructor as the `ModelH264`, and add a
```err = myMod.AddModelFromRegistry(ctx, camera.API, viamrtsp.Model)```

It's a little off kilter from how we currently register different models in one module but should work.

File: Makefile (hunk 5 above)
Comment (seanavery) [reply Makefile:41]: Good call! We do not need an AppImage. 
- Tested that packaging FFmpeg libs, go binary, and shell entry works for Linux.
- Going to try out a static FFmpeg build/link so that we can just ship the go binary.

File: rtsp.go [hunk 6]
```diff
@@ -164,47 +194,54 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
 	if err != nil {
 		return err
 	}
 
-	// if SPS and PPS are present into the SDP, send them to the decoder
+	// For H.264, handle SPS and PPS (assuming your H264 format has these fields)
 	if format.SPS != nil {
 		rc.rawDecoder.decode(format.SPS)
+	} else {
+		rc.logger.Warn("no SPS found in H264 format")
```
nicksanford, line 211: Is it possible for h264 -> JPEG decoding to succeed if `format.SPS == nil` and / or `format.PPS == nil`

rtsp.go
```diff
@@ -164,47 +194,54 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
+				rc.logger.Warn("waiting for I-frame")
+				return
+			}
+			iFrameReceived = true
+		}
 
+		for _, nalu := range au {
 			if len(nalu) < 20 {
-				// TODO(ERH): this is probably wrong, but fixes a spam issue with "no frame!"
+				// TODO: this is probably wrong, but fixes a spam issue with "no frame!"
```
nicksanford, line 241: What exactly is the issue this is trying to prevent?

rtsp.go
```diff
@@ -164,47 +194,54 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
+		rc.logger.Warn("no PPS found in H264 format")
 	}
 
+	iFrameReceived := false
+
 	// On packet retreival, turn it into an image, and store it in shared memory
 	rc.client.OnPacketRTP(track, format, func(pkt *rtp.Packet) {
-		// extract access units from RTP packets
-		au, _, err := rtpDec.Decode(pkt)
+		au, _, err := rtpDec.DecodeUntilMarker(pkt)
```
nicksanford, line 223: Why are we using `DecodeUntilMarker` as opposed to `Decode`?

rtsp.go
```diff
@@ -27,18 +31,17 @@ import (
 	"go.viam.com/rdk/resource"
 )
 
+/*
+#include <libavcodec/avcodec.h>
+*/
+import "C"
```
nicksanford, line 37: Could I recommend keeping this file pure go & having all CGO-specific code contained in the same place (like decoder.go or maybe a new file call codec.go)?

rtsp.go
```diff
@@ -164,47 +194,54 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
-	// setup RTP/H264 -> H264 decoder
 	rtpDec, err := format.CreateDecoder2()
 	if err != nil {
-		return err
+		rc.logger.Errorf("error creating H264 decoder %v", err)
 	}
 
-	// setup H264 -> raw frames decoder
-	rc.rawDecoder, err = newH264Decoder()
+	rc.rawDecoder, err = newDecoder(C.AV_CODEC_ID_H264)
```
nicksanford, line 202: How about `newH264Decoder` so that the caller doesn't need to know about libav?

rtsp.go
```diff
@@ -215,16 +252,102 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
 	}
-	clientSuccessful = true
+
+	rtpDec, err := format.CreateDecoder2()
+	if err != nil {
+		rc.logger.Errorf("error creating H265 decoder %v", err)
+		return err
+	}
+
+	rc.rawDecoder, err = newDecoder(C.AV_CODEC_ID_HEVC)
```
nicksanford, line 283: How about `newH265Decoder` so that the caller doesn't need to depend on libav?

decoder.go
```diff
@@ -20,18 +32,45 @@ func frameLineSize(frame *C.AVFrame) *C.int {
...
 
-// h264Decoder is a wrapper around ffmpeg's H264 decoder.
-type h264Decoder struct {
-	codecCtx    *C.AVCodecContext
-	srcFrame    *C.AVFrame
-	swsCtx      *C.struct_SwsContext
-	dstFrame    *C.AVFrame
-	dstFramePtr []uint8
+// getStreamInfo opens a stream URL and retrieves its information
+func getStreamInfo(url string) string {
```
nicksanford, line 36: Can we please get some definition of what the expected behavior is when a url supports both H264 & H265? 
Is one always preferred over the other? 
Is it nondeterministic?
Are there any product requirements we have been given regarding how this should work?

decoder.go
```diff
@@ -20,18 +32,45 @@ func frameLineSize(frame *C.AVFrame) *C.int {
...
+	}
+	defer C.avformat_close_input(&avFormatCtx)
+
+	// Try to retrieve the stream information
+	if C.avformat_find_stream_info(avFormatCtx, nil) < 0 {
+		return "Error: Could not retrieve the stream information"
+	}
+
+	// Iterate over all streams and check codec ID
+	for i := C.uint(0); i < C.uint(avFormatCtx.nb_streams); i++ {
```
nicksanford, line 59: Are we sure that go i++ does the right thing with C `uint`s? 
Could I recommend instead:
```
for i := 0; i < int(avFormatCtx.nb_streams); i++ {
...
}
```

decoder.go
```diff
@@ -73,7 +112,8 @@ func (d *h264Decoder) close() {
 	C.avcodec_close(d.codecCtx)
 }
 
-func (d *h264Decoder) decode(nalu []byte) (image.Image, error) {
+func (d *decoder) decode(nalu []byte) (image.Image, error) {
+	// 0x00000001
```
nicksanford, line 116: what is meant by this comment?

etc/rtsp-simple-server.yml
```diff
@@ -0,0 +1,4 @@
+protocols: [tcp]
```
nicksanford, line 1: How is this intended to be used?

go.mod
```diff
@@ -163,3 +163,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	nhooyr.io/websocket v1.8.7 // indirect
 )
+
+replace go.viam.com/rdk => github.com/abe-winter/rdk v0.0.0-20240220221954-caf1960a7316
```
nicksanford, line 167: Is this intended to still be here?

rtsp.go
```diff
@@ -164,47 +194,54 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
+		au, _, err := rtpDec.DecodeUntilMarker(pkt)
 		if err != nil {
 			if err != rtph264.ErrNonStartingPacketAndNoPrevious && err != rtph264.ErrMorePacketsNeeded {
 				rc.logger.Errorf("error decoding(1) h264 rstp stream %v", err)
 			}
 			return
 		}
 
-		for _, nalu := range au {
+		if !iFrameReceived {
```
nicksanford, line 231: Is this change to fix a specific problem?

PR #2, opened 2024-03-15
RSDK-6814 - Add Codec Detection and H265 Support

seanavery:
## Description

Unlike the [first PR](https://github.com/erh/viamrtsp/pull/1) which contains development packaging work, this aims to be the minimal change needed to add Codec Detection and H265 support. Build work can be introduced in subsequent PRs.

- This PR allows the module binary to handle both H264 and H265 codecs at runtime.
- If a given endpoint has tracks that support both codecs, H264 will take preference.

### Build/Run on Linux:
- Make sure FFmpeg deps are installed: 
`sudo apt install libswscale-dev libavcodec-dev libavformat-dev libavutil-dev`
- Build go binary: 
`make bin/viamrtsp`
- Run as "local" module on Linux machine with FFmpeg deps installed.
```
    {
      "executable_path": "/path/to/viamrtsp",
      "name": "viamrtsp",
      "type": "local"
    }
```
- Configure component to listen to RTSP server endpoint.
```
    {
      "model": "erh:viamrtsp:rtsp-h264",
      "name": "ip-cam",
      "namespace": "rdk",
      "type": "camera",
      "attributes": {
        "rtsp_address": "rtsp://localhost:8554/live.stream"
      },
      "depends_on": []
    }
```
### Create Fake RTSP Camera:
- Download mediamtx:
`wget https://github.com/bluenviron/mediamtx/releases/download/v1.6.0/mediamtx_v1.6.0_linux_arm64v8.tar.gz`
- Extract:
`tar -xzf mediamtx_v1.6.0_linux_arm64v8.tar.gz`
- Run:
`./mediamtx`
- Run fake RTSP camera script:
`ffmpeg -re -f lavfi -i testsrc=size=640x480:rate=30 -vcodec libx264 -pix_fmt yuv420p -f rtsp -rtsp_transport tcp rtsp://0.0.0.0:8554/live.stream`
  - Replace codec with `libx265` for H265 feed.

## Tests
- Tested on RPI4 Bullseye (with FFmpeg dependencies installed).
    - fmpeg version `4.3.5`
- Tested on Macos.
    - ffmpeg version `6.1.1`
- `H264`, `H265`, `H264H`, and `H264B` streams working w/ physical IP Camera.
    - **Warning:** There is a known issue where toggling between h264h and h264b/h264 causes garbled video from the decoder. This PR does not address this.
- Endpoint without supported codec returns Unknown from getStreamInfo helper and prints error log correctly.
- avError fetches error messages correctly.
[...]

PR #4, opened 2024-03-21
Add Config struct to rtsp module

randhid:
We're removing builtin/rtsp, but this module has a dependency on that package. The only thing it's doing is pulling the Config Struct and Validate function from that package. I move them here.


cmd/remote/cmd.go
```diff
@@ -26,7 +25,7 @@ func main() {
 func realMain() error {
 
 	ctx := context.Background()
-	logger := golog.NewDevelopmentLogger("client")
+	logger := logging.NewLogger("client")
```
randhid, line 28: The module would not build locally unless I changed the logger to the new  style logger in the rdk. runWebWithConfig needed a `logging.Logger` for me. Can rever if this is not the case on whatever this is being built on.

rtsp.go
```diff
@@ -21,19 +21,19 @@ import (
 	goutils "go.viam.com/utils"
 
 	"go.viam.com/rdk/components/camera"
-	"go.viam.com/rdk/components/camera/rtsp"
```
randhid, line 24: This package will eventually go away. Config used to be : https://github.com/viamrobotics/rdk/blob/0a7f6b23851f8c04f82ffa5583f600668e5e2e3a/components/camera/rtsp/rtsp.go#L46

randhid:
@erh - this removed the dependency of this module on the builtin rtsp camera go package.

cc @seanavery @nicksanford

erh:
can you also upgrade to newest rdk as part of this.

cmd/remote/cmd.go
nicksanford: Changing to `logging.NewLogger` will no longer log debug-level logs (which the previous logger did).

From a cursory glance, the only code in this repo that logs at debug level is here: https://github.com/erh/viamrtsp/blob/main/rtsp.go#L128

Could I recommend changing this to:
```suggestion
	logger := module.NewLoggerFromArgs("client")
```

That way module will default to info level logs but allow the end user to configure the level at runtime: https://pkg.go.dev/go.viam.com/rdk@v0.23.2/module#NewLoggerFromArgs

nicksanford (approved):
Looks good minus nit & updating to latest RDK

cmd/remote/cmd.go
randhid: Thanks! I'll make the change.

nicksanford:
@randhid do you not also need to remove this line from cmd/remote/main.go: https://github.com/erh/viamrtsp/pull/4/files#diff-788d8d6232a7bfca6baf42aeb2f5334425bade25a2d8ee4ca3ba6e89cb2fc7afR8?

randhid:
> @randhid do you not also need to remove this line from cmd/remote/main.go: https://github.com/erh/viamrtsp/pull/4/files#diff-788d8d6232a7bfca6baf42aeb2f5334425bade25a2d8ee4ca3ba6e89cb2fc7afR8?

yep, thanks for catching that.

.canon.yaml
```diff
@@ -0,0 +1,8 @@
+rtsp:
```
randhid, line 1: @nicksanford is this useful for you and sean or should I add it to the git ignore?

nicksanford: I don't mind it :) 

It can stay from my perspective

.canon.yaml
```diff
@@ -0,0 +1,8 @@
...
+    image_amd64: ghcr.io/viamrobotics/canon:amd64
+    image_arm64: ghcr.io/viamrobotics/canon:arm64
+    minimum_date: 2022-12-20T18:49:25.781750658Z
+    update_interval: 168h0m0s
+    persistent: false
+    user: testbot
+    group: testbot
```
randhid, line 8:
```suggestion
    group: testbot
    
```

.canon.yaml
seanavery: Canon wrapper is useful - using `antique` containers for static build PR.

seanavery (commented):
LGTM!
- Will rebase [Codec Detection & H265 PR](https://github.com/erh/viamrtsp/pull/2) with this change.

.canon.yaml
randhid: I'll add that!

.gitignore
```diff
@@ -26,3 +26,4 @@ go.work
 
 viamrtsp
 module.tar.gz
+
```
erh, line 29: remove

erh (approved):
minus that one nit

.gitignore
randhid: removed.

PR #5, opened 2024-03-25
RSDK-6974 - Add FFmpeg Static Build to Makefile

seanavery:
## Description

This PR adds a Makefile script to clone and build `FFmpeg v6.1` static libraries and CGO flags for linking to the go binary.

Here are the remaining dynamic links:
```
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libz.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libresolv.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```
- Have verified that these libraries are available across Ubuntu and Debian versions.
- Building module in the antique container allows for libc support `>=2.27`.
- The size change is pretty small due to disabling all but necessary FFmpeg features.
    - Size could be reduced further by stripping symbols if we want 
```
47 MB -> 55 MB
```

## Try Out

To build for linux/arm64 follow these steps:
- Startup canon:
`canon -arch arm64`
- Build FFmpeg and go binary:
`make bin/viamrtsp`

## Tests
- FFmpeg static build/link working inside arm64/amd64 canon.
- FFmpeg build and go build still working on MacOS.
- Go build linking against FFmpeg static libs successfully through the package config flag.
- Tested binary works on:
    - Pi4 Debian Bullseye :white_check_mark:
    - Pi5 Debian Bookworm :white_check_mark:
    - Jetson Nano Ubuntu 18.04 :white_check_mark:
    - Jetson Orin Nano Ubuntu 20.04 :white_check_mark:
 - Fixed log spam issue an have tested that decoders working the same as with dynamic link.
   - H264 physical ip camera stream ✅ 
   - H265 physical ip camera stream ✅ 

## Questions
- Are we ok with using canon for Linux builds? 
    - This PR assumes libjpeg static lib is available for fully functional (standalone) binary.
        - If on linux, use archive file, else dynamic link will be created.
- Are we ok with not having a fully static build for MacOS? 
    - Builds work for testing but will not be able to statically link to libjpeg.
- Are we happy with using FFmpeg v6.1? 
    - There are no API changes between 4.x and 6.x but looks to be lots of optimizations made. https://www.ffmpeg.org/index.html#pr6.1

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "FFmpeg"]
+	path = FFmpeg
+	url = https://github.com/FFmpeg/FFmpeg.git
+	branch = release/6.1
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "FFmpeg"]
-	path = FFmpeg
-	url = https://github.com/FFmpeg/FFmpeg.git
-	branch = tags/n6.1
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "FFmpeg"]
 	path = FFmpeg
 	url = https://github.com/FFmpeg/FFmpeg.git
-	branch = release/6.1
+	branch = tags/n6.1
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ FFmpeg:
 	git submodule init && git submodule update
 
 $(FFMPEG_PREFIX): FFmpeg
-	cd FFmpeg && ./configure $(FFMPEG_OPTS) && make -j$(shell nproc) && make install
+	cd FFmpeg && ./configure $(FFMPEG_OPTS) && $(MAKE) -j$(shell nproc) && $(MAKE) install
 
 build-ffmpeg:
 ifeq ($(UNAME_S),Linux)
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ updaterdk:
 	go mod tidy
 
 FFmpeg:
-	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch release/6.1
+	git submodule init && git submodule update
 
 $(FFMPEG_PREFIX): FFmpeg
 	cd FFmpeg && ./configure $(FFMPEG_OPTS) && make -j$(shell nproc) && make install
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --disable-everything \
                --enable-decoder=h264 \
                --enable-decoder=hevc \
+               --enable-network \
+               --enable-protocol=http \
                --enable-swscale
 CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
 ifeq ($(UNAME_S),Linux)
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ endif
 
 bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
 	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
+		CGO_LDFLAGS=$(CGO_LDFLAGS) \
 		go build -o bin/viamrtsp cmd/module/cmd.go
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,10 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --enable-decoder=h264 \
                --enable-decoder=hevc \
                --enable-network \
-               --enable-demuxer=rtsp
+			   --enable-demuxer=rtsp \
+			   --enable-parser=h264 \
+			   --enable-parser=hevc
+
 CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
 ifeq ($(UNAME_S),Linux)
 	CGO_LDFLAGS := "$(CGO_LDFLAGS) -l:libjpeg.a"
--- a/Makefile
+++ b/Makefile
@@ -12,17 +12,13 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --enable-swscale
 CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
 ifeq ($(UNAME_S),Linux)
-ifneq ($(shell find / -name libjpeg.a 2> /dev/null),)
 	CGO_LDFLAGS := "$(CGO_LDFLAGS) -l:libjpeg.a"
-endif
 endif
 
 .PHONY: build-ffmpeg test lint updaterdk module
 
 bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
 	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
-		CGO_CFLAGS=-I$(FFMPEG_PREFIX)/include \
-		CGO_LDFLAGS="-L$(FFMPEG_PREFIX)/lib -l:libjpeg.a" \
 		go build -o bin/viamrtsp cmd/module/cmd.go
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --enable-decoder=h264 \
                --enable-decoder=hevc \
                --enable-network \
-               --enable-protocol=http \
+               --enable-demuxer=rtsp \
                --enable-swscale
 CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
 ifeq ($(UNAME_S),Linux)
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq ($(UNAME_S),Linux)
 	CGO_LDFLAGS := "$(CGO_LDFLAGS) -l:libjpeg.a"
 endif
 
-.PHONY: build-ffmpeg test lint updaterdk module
+.PHONY: build-ffmpeg test lint updaterdk module clean
 
 bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
 	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 UNAME_S ?= $(shell uname -s)
 UNAME_M ?= $(shell uname -m)
-FFMPEG_PREFIX ?= $(shell pwd)/ffmpeg/$(UNAME_S)-$(UNAME_M)
+FFMPEG_PREFIX ?= $(shell pwd)/FFmpeg/$(UNAME_S)-$(UNAME_M)
 FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --enable-static \
                --disable-shared \
@@ -33,7 +33,7 @@ updaterdk:
 	go mod tidy
 
 FFmpeg:
-	git submodule init && git submodule update
+	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch n6.1
 
 $(FFMPEG_PREFIX): FFmpeg
 	cd FFmpeg && ./configure $(FFMPEG_OPTS) && $(MAKE) -j$(shell nproc) && $(MAKE) install
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,7 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
                --enable-decoder=h264 \
                --enable-decoder=hevc \
                --enable-network \
-               --enable-demuxer=rtsp \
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -1,6 +1,29 @@
...
+               --disable-programs \
+               --disable-doc \
+               --disable-everything \
+               --enable-decoder=h264 \
+               --enable-decoder=hevc \
+               --enable-swscale
+CGO_LDFLAGS := -L$(FFMPEG_PREFIX)/lib
+ifeq ($(UNAME_S),Linux)
+ifneq ($(shell find / -name libjpeg.a 2> /dev/null),)
+	CGO_LDFLAGS := "$(CGO_LDFLAGS) -l:libjpeg.a"
```
seanavery, line 16: Only statically links `libjpeg.a` if found on system.

Makefile
```diff
@@ -1,6 +1,29 @@
...
+endif
+endif
 
-bin/viamrtsp: *.go cmd/module/*.go
-	go build -o bin/viamrtsp cmd/module/cmd.go
+.PHONY: build-ffmpeg test lint updaterdk module
+
+bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
+	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
```
seanavery, line 23: Manually add package config path here to point to ffmpeg build.
(Had to add this when moving to antique)

Makefile
```diff
@@ -12,5 +35,19 @@ updaterdk:
 	go get go.viam.com/rdk@latest
 	go mod tidy
 
+FFmpeg:
+	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch release/6.1
+
+$(FFMPEG_PREFIX): FFmpeg
```
seanavery, line 41: This will prevent slowdowns from configure calls if install path already exists.

Makefile
```diff
@@ -12,5 +35,19 @@ updaterdk:
...
+	cd FFmpeg && ./configure $(FFMPEG_OPTS) && make -j$(shell nproc) && make install
+
+build-ffmpeg:
+ifeq ($(UNAME_S),Linux)
+ifeq ($(UNAME_M),x86_64)
+	which nasm || (sudo apt update && sudo apt install -y nasm)
```
seanavery, line 47: In amd64 antique container, we need to install nasm for the optimized assembly kernels to work.

README.md
```diff
@@ -8,6 +8,10 @@ Notes
 ===
 * Heavily cribbed from https://github.com/bluenviron/gortsplib/blob/main/examples/client-read-format-h264-convert-to-jpeg/main.go
 
+FFmpeg
```
nicksanford, line 11: Do we need to change the readme instructions under `Prep linux` to say that the apt packages must only be installed on the build machine but that the prod machine doesn't need to have those packages?

Makefile
nicksanford: Why only if found? Isn't it a requirement? 
Also, is this the recommended way to find a `.a` file? isn't pkg-config the common way to find C/C++ deps on linux?

Makefile
```diff
@@ -1,6 +1,29 @@
...
+endif
 
-bin/viamrtsp: *.go cmd/module/*.go
-	go build -o bin/viamrtsp cmd/module/cmd.go
+.PHONY: build-ffmpeg test lint updaterdk module
+
+bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
+	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
+		CGO_CFLAGS=-I$(FFMPEG_PREFIX)/include \
+		CGO_LDFLAGS="-L$(FFMPEG_PREFIX)/lib -l:libjpeg.a" \
```
nicksanford, line 25: Isn't this going to overwrite `CGO_LDFLAGS` here: https://github.com/erh/viamrtsp/pull/5/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R13 and here https://github.com/erh/viamrtsp/pull/5/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R16

?

Makefile
nicksanford: 1. How do we know if a linux x86_64 executable of this module will use the optimized assembly kernels?

2. Does this put any requirements on the machine that runs the executable (libraries, OS version, kernel version, etc)?

README.md
```diff
@@ -8,6 +8,10 @@ Notes
...
+===
+
+This project is designed to work with FFmpeg version 6.1. During the build process of the module, platform-specific static builds of FFmpeg 6.1 will be created. This ensures that the module is always using the correct version of FFmpeg, regardless of what other versions might be installed on your system.
```
nicksanford, line 14: Can we please add instructions for how you recommend production builds of the module be produced on linux x86_64, linux arm, and macos arm?

Makefile
```diff
@@ -12,5 +35,19 @@ updaterdk:
 	go get go.viam.com/rdk@latest
 	go mod tidy
 
+FFmpeg:
+	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch release/6.1
```
nicksanford, line 39: I like this but could we make ffmpeg a git submodule? That way we can keep track of which specific commit of FFmpeg we are building against throughout the commit history of this module.

Here is an example of how we did that in viam-cartographer:
https://github.com/viamrobotics/viam-cartographer/tree/main/viam-cartographer

https://github.com/viamrobotics/viam-cartographer/blob/main/Makefile#L51

LMK if you would like to pair on this

Makefile
seanavery: - Added that check in so that build will work on macos or linux without the static build
    - apt package: https://libjpeg-turbo.org/Downloads/YUM
    - macos uses `jpeg-turbo` homebrew forumula. https://formulae.brew.sh/formula/jpeg-turbo 
        - `.a` actually available on macos
```
ls /opt/homebrew/Cellar/jpeg-turbo/3.0.1/lib 
cmake                    libjpeg.dylib            libturbojpeg.dylib
libjpeg.8.3.2.dylib      libturbojpeg.0.3.0.dylib pkgconfig
libjpeg.8.dylib          libturbojpeg.0.dylib
libjpeg.a                libturbojpeg.a
```
- `pc` file links to the shared object, so have to specify link to the `.a`.
```
cat /usr/lib/x86_64-linux-gnu/pkgconfig/libjpeg.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libjpeg
Description: JPEG image codec
Version: 
Libs: -L${libdir} -ljpeg
Cflags: -I${includedir}
```
- Do not like this find check, it is really slow on macos...
    - I think we can make the assumption that .a is available since we get it on macos as well.
    - Will refactor this so libjpeg static link also works on macos and find check is removed.

Makefile
seanavery: - ASM code should just be compiled in the library
- For example, check out h264 asm backend here: https://github.com/FFmpeg/FFmpeg/blob/8ca57fcf9ed327a6c2d9c5345be9b7e0724ca048/libavcodec/x86/h264_deblock.asm#L4

- Do not think there are additional considerations for kernel/os for asm stuff.
    - Any system calls should have the same requirements as from the go/c/cpp code.
- Just needed the `nasm` utility to build the asm code through ffmpeg builder.
        - If it cannot build assembly it will default to the unoptimized c code, so that it was important to include.

README.md
seanavery: Good call! We no longer need apt installs. The build system expects local FFmpeg build. Edited this section with build instructions and a note on static build.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "FFmpeg"]
+	path = FFmpeg
+	url = https://github.com/FFmpeg/FFmpeg.git
+	branch = release/6.1
```
nicksanford, line 4: FYI The submodule module commit you commited is this one:
https://github.com/FFmpeg/FFmpeg/commits/f309408874a5d6ccc77063631fa78b8fb59937a0/

that is different from the 6.1 release commit: https://github.com/FFmpeg/FFmpeg/commits/n6.1/

Can you please update it?

seanavery: Thanks, moved to `tags/n6.1` which points to the commit `d4ff0020b40b524a490cf62eccbd3a318f4c0e58`.

README.md
seanavery: Added in instructions for arm64/x86 builds.
- For now, will not be able to have a fully static build for macos
    - Requires libjpegturbo dylib on device (Which seems to be the case for viam-server as well).

Makefile
seanavery: Ok, currently:
- On linux, static link to libjpeg.
- On macos, dynamic link to libjpeg.
- FFmpeg static link works on both.

randhid:
@erh this is ready for you to review or merge.

nicksanford:
Running `make bin/viamrtsp` on osx fails with this error:
```bash
make bin/viamrtsp
/Library/Developer/CommandLineTools/usr/bin/make /Users/nicksanford/code/viamrtsp/ffmpeg/Darwin-arm64
cd FFmpeg && ./configure --prefix=/Users/nicksanford/code/viamrtsp/ffmpeg/Darwin-arm64 --enable-static --disable-shared --disable-programs --disable-doc --disable-everything --enable-decoder=h264 --enable-decoder=hevc --enable-swscale && /Library/Developer/CommandLineTools/usr/bin/make -j8 && /Library/Developer/CommandLineTools/usr/bin/make install
/bin/sh: ./configure: No such file or directory
make[1]: *** [/Users/nicksanford/code/viamrtsp/ffmpeg/Darwin-arm64] Error 127
make: *** [build-ffmpeg] Error 2
```

nicksanford (changes requested):
runinng `make make bin/viamrtsp` from a clean slate doesn't succeed on either osx or arm64 cannon. 

Please let me know when it's fixed & I can re-test.

seanavery:
@nicksanford There was an issue with the `FFmpeg` path check when clone will initialize empty folder for submodule.

I ended up refactoring back to use make target instead of submodule as it was much cleaner.

- I think this accomplishes anchoring to a commit as we are using a release tag branch.
- Instead of writing shell scripts to check if files exist, it seems easier to use Makefile builtin target check. Especially since I am chaining the Install path (FFMPEG_PREFIX) and FFmpeg clone path to avoid re-running these tasks.

seanavery:
Ok @nicksanford , passing CI now. See run [here](https://github.com/erh/viamrtsp/actions/runs/8545154527/job/23412784960). Had to add in rtsp-demuxer to handle open for codec detection. We can remove this once that call is removed.

nicksanford:
@seanavery initial build on osx succeeded. I'll test it more tomorrow morning & approve if I don't find any more issues. Thanks!

nicksanford:
@seanavery
A few issues:
1. 
on OSX, when I run the module I get the following spam over & over again
```
\_ [h264 @ 0x125610c40] Error splitting the input into NAL units.
2024-04-04T10:09:14.777-0400    ERROR   robot_server.process.viamrtsp_/Users/nicksanford/code/viamrtsp/bin/viamrtsp.StdErr      pexec/managed_process.go:252
\_ [h264 @ 0x125610c40] No start code is found.
2024-04-04T10:09:14.777-0400    ERROR   
```

I do not get that spam output on main

2. Running `make` on osx followed by canon result in the canon build failing, probably b/c it is reusing the build directory from the different platform. The solution was to manually delete the FFmpeg directory and re-run `make`. Could you please add a `make clean` command that leaves the build in a clean state?

nicksanford (changes requested):
There are spam warnings on OSX & running the build on osx followed by canon doesn't work.

Please retest on the following platforms & confirm that the executable works & doesn't spam warnings:
1. pi 4
2. pi 5 
3. x86_64 linux
4. arm mac

seanavery:
@nicksanford 
- Thanks for the comment about clean script, added!
- I see extra logs on decoder from macos static build.
    - I think decoder logs should be handled separately with a helper to change ffmpeg loglevel.
    - Tested on my mac and the pipeline is still working well.
 - Confirmed that going from darwin/arm64 -> canon linux/arm64 causes build issues
     - This does not happen when going to x86 container since it will rebuild binaries.
     - Hopefully, the clean script is sufficient for now.
  - Retested:
      - macos arm64 ✅ 
      - pi4 ✅ 
      - pi5 ✅ 
      - x86 ✅

nicksanford (approved):
Looks good to me.

I tested it on:
MacOS aarm64: `14.4.1`
Ubuntu X86_64:  `22.04.4 LTS`
Raspberry Pi 5 arm64: Raspbian Lite: Bookworm
Raspberry Pi 4 arm64: Raspbian Lite: Bullseye

README.md
```diff
@@ -1,14 +1,18 @@
 
-Prep linux
+Build for Linux
```
nicksanford, line 2: [nit] Mind please also adding osx instructions? Should just be `make clean; make` right?

PR #6, opened 2024-03-27
[RSDK-7079] Refactor models

hexbabe:
## Context

https://viam.atlassian.net/browse/RSDK-7079

Adds `rtsp` model that streams from any available, supported codec

Makes `rtsp-h264` return error if detected codec is not h264

## Question
If we're adding a generic `rtsp` model, but also keeping `rtsp-h264`, should we add `rtsp-h265`? Plus more accordingly to new codecs we introduce?

Edit: I went ahead and added `rtsp-h265` as well— can remove it if bad

## Local Testing

1. `h264` fake stream with model `rtsp`: Success ✅
2. `h264` fake stream with model `rtsp-h264`: Success ✅
3. `h264` fake stream with model `rtsp-h265`: `resource build error: rpc error: code = Unknown desc = h265 track not found` ✅

5. `h265` fake stream with model `rtsp`: Success ✅
6. `h265` fake stream with model `rtsp-h264`: `resource build error: rpc error: code = Unknown desc = h264 track not found` ✅
7. `h265` fake stream with model `rtsp-h265`: Success ✅

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -34,10 +34,18 @@ var ModelAgnostic = family.WithModel("rtsp")
 var ModelH264 = family.WithModel("rtsp-h264")
 var ModelH265 = family.WithModel("rtsp-h265")
 var Models = []resource.Model{ModelAgnostic, ModelH264, ModelH265}
-var modelToCodecMap = map[resource.Model]videoCodec{
-	ModelAgnostic: Unknown,
-	ModelH264:     H264,
-	ModelH265:     H265,
+
+func modelToCodec(model resource.Model) videoCodec {
+	switch model {
+	case ModelAgnostic:
+		return Unknown
+	case ModelH264:
+		return H264
+	case ModelH265:
+		return H265
+	default:
+		return Unknown
+	}
 }
 
 func init() {
@@ -377,7 +385,7 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 		u:      u,
 		logger: logger,
 	}
-	codecInfo := modelToCodecMap[conf.Model]
+	codecInfo := modelToCodec(conf.Model)
 	err = rtspCam.reconnectClient(codecInfo)
 	if err != nil {
 		return nil, err
--- a/rtsp.go
+++ b/rtsp.go
@@ -30,12 +30,16 @@ import (
 )
 
 var family = resource.ModelNamespace("erh").WithFamily("viamrtsp")
+var Model = family.WithModel("rtsp")
 var ModelH264 = family.WithModel("rtsp-h264")
+var Models = []resource.Model{Model, ModelH264}
 
 func init() {
-	resource.RegisterComponent(camera.API, ModelH264, resource.Registration[camera.Camera, *Config]{
-		Constructor: newRTSPCamera,
-	})
+	for _, model := range Models {
+		resource.RegisterComponent(camera.API, model, resource.Registration[camera.Camera, *Config]{
+			Constructor: newRTSPCamera,
+		})
+	}
 }
 
 // Config are the config attributes for an RTSP camera model.
@@ -78,6 +82,8 @@ type rtspCamera struct {
 	activeBackgroundWorkers sync.WaitGroup
 
 	latestFrame atomic.Pointer[image.Image]
+
+	mustUseH264 bool
 
 	logger logging.Logger
 }
@@ -173,6 +179,9 @@ func (rc *rtspCamera) reconnectClient() (err error) {
 	codecInfo, err := getStreamInfo(rc.u.String())
 	if err != nil {
 		return err
+	}
+	if rc.mustUseH264 && codecInfo != H264 {
+		return errors.New("h264 requested, but h264 codec not detected")
 	}
 
 	switch codecInfo {
@@ -362,8 +371,9 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 		return nil, err
 	}
 	rtspCam := &rtspCamera{
-		u:      u,
-		logger: logger,
+		u:           u,
+		logger:      logger,
+		mustUseH264: conf.Model == ModelH264,
 	}
 	err = rtspCam.reconnectClient()
 	if err != nil {
--- a/rtsp.go
+++ b/rtsp.go
@@ -2,6 +2,7 @@ package viamrtsp
 
 import (
 	"context"
+	"fmt"
 	"image"
 	"io"
 	"sync"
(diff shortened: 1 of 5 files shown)


rtsp.go
```diff
@@ -79,6 +83,8 @@ type rtspCamera struct {
 
 	latestFrame atomic.Pointer[image.Image]
 
+	mustUseH264 bool
```
randhid, line 86: In the future we can pass in the codecInfo for multiple encodings to `reconnectClient` and have it fail if it's not the same codec, what you've done is fine for now, also lets us not keep track of as much state in this struct.

hexbabe: > pass in the codecInfo for multiple encodings to reconnectClient

But isn't `codecInfo` handled inside `reconnectClient`

seanavery: Yes agree, would be nice to not add to `rtspCamera` struct here, but this seems good for now since we only have two models.

For "stateless" option, we could:
- Add codec whitelist param to `reconnectClient`.
- Pass codec whitelist to `clientReconnectBackgroundWorker` based on model.

hexbabe: Would it be Viam `Act Decisively` of me to just go ahead and add model `rtsp-h265` and implement a more elegant solution with enums.

hexbabe: I'm gonna do it lol

rtsp.go
```diff
@@ -138,7 +148,7 @@ func (rc *rtspCamera) closeConnection() error {
 }
 
 // reconnectClient reconnects the RTSP client to the streaming server by closing the old one and starting a new one.
-func (rc *rtspCamera) reconnectClient() (err error) {
+func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
```
hexbabe, line 151: Nit to consider:

Is passing around `codecInfo` worth omitting it from state? Will we need to access `codecInfo` in separate branches of logic?

rtsp.go
```diff
@@ -30,9 +30,15 @@ import (
...
+var ModelAgnostic = family.WithModel("rtsp")
 var ModelH264 = family.WithModel("rtsp-h264")
-var Models = []resource.Model{Model, ModelH264}
+var ModelH265 = family.WithModel("rtsp-h265")
+var Models = []resource.Model{ModelAgnostic, ModelH264, ModelH265}
+var modelToCodecMap = map[resource.Model]videoCodec{
+	ModelAgnostic: Unknown,
+	ModelH264:     H264,
+	ModelH265:     H265,
+}
```
randhid, line 41: Worth looking at using a `iota` 
https://yourbasic.org/golang/iota/

hexbabe: I think `Unknown` is an iota

rtsp.go
```diff
@@ -170,9 +180,11 @@ func (rc *rtspCamera) reconnectClient() (err error) {
 		return err
 	}
 
-	codecInfo, err := getStreamInfo(rc.u.String())
-	if err != nil {
-		return err
+	if codecInfo == Unknown {
```
seanavery, line 183: Nice, I like how we do not have to modify decoder.go getStreamInfo helper at all, and can just lookup if when using the agnostic model.

rtsp.go
seanavery: Either way makes sense to me.

randhid: same. if you'd like to have less state go ahead.

rtsp.go
hexbabe: Yeah good point especially if in the future we want more than a direct mapping in our model <-> codec relationship. Added change in new commit

randhid (approved):

randhid:
@erh this is ready for you to review or merge.

rtsp.go
```diff
@@ -30,12 +30,30 @@ import (
...
+func modelToCodec(model resource.Model) videoCodec {
+	switch model {
+	case ModelAgnostic:
+		return Unknown
+	case ModelH264:
+		return H264
+	case ModelH265:
+		return H265
+	default:
+		return Unknown
```
nicksanford, line 47: I think this case should be an error. So you will want to change the signature of this function to return an `(videoCodec, err)`

nicksanford, line 47:
```suggestion
		return Unknown, errors.New("model codec not specified")
```

nicksanford (changes requested):

rtsp.go
hexbabe: addressed

hexbabe: done

rtsp.go
```diff
@@ -30,12 +30,30 @@ import (
...
 var family = resource.ModelNamespace("erh").WithFamily("viamrtsp")
+var ModelAgnostic = family.WithModel("rtsp")
 var ModelH264 = family.WithModel("rtsp-h264")
+var ModelH265 = family.WithModel("rtsp-h265")
+var Models = []resource.Model{ModelAgnostic, ModelH264, ModelH265}
+
+func modelToCodec(model resource.Model) videoCodec {
+	switch model {
+	case ModelAgnostic:
+		return Unknown
```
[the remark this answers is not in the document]
hexbabe: done

rtsp.go
```diff
@@ -30,12 +30,30 @@ import (
 )
 
 var family = resource.ModelNamespace("erh").WithFamily("viamrtsp")
+var ModelAgnostic = family.WithModel("rtsp")
 var ModelH264 = family.WithModel("rtsp-h264")
+var ModelH265 = family.WithModel("rtsp-h265")
+var Models = []resource.Model{ModelAgnostic, ModelH264, ModelH265}
+
+func modelToCodec(model resource.Model) videoCodec {
```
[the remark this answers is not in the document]
hexbabe: done

randhid (approved):

rtsp.go
```diff
@@ -365,7 +373,11 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 		u:      u,
 		logger: logger,
 	}
-	err = rtspCam.reconnectClient()
+	codecInfo, err := modelToCodec(conf.Model)
```
nicksanford, line 376: Can we please change `codecInfo` to be a struct member on `rtspCamera` rather than a perameter given that it never changes for the lifetime of an `rtspCamera` struct?

randhid: I'd actually rather pass it around in the creator/reconfigure since it's only used there than have it on the struct, it's less state to check in other API calls in this driver.

nicksanford (approved):

seanavery (approved):
LGTM

PR #3, opened 2024-03-19
[IGNORE] - Nick test passthrough

nicksanford:
# NOT READY YET

For: https://github.com/viamrobotics/rdk/pull/3708

PR #7, opened 2024-03-27
[RSDK-1155] experimental rtp_passthrough

nicksanford:
Based on: https://github.com/erh/viamrtsp/pull/3

Depends on viam-server / RDK changes: 
https://github.com/viamrobotics/rdk/pull/3773
https://github.com/erh/viamrtsp/pull/11

Changes:
1. Add `rtp_passthrough` config parameter. Only enable H264 passthrough features if set to true
1. Implement h264 RTP passthrough features.
1. change `cmd/module/cmd.go` to use `utils.ContextualMain` so RTSP connections get closed when the module process  is sent kill / quit signals
1. Disable all LibAV logs other than fatal ones by default. If the module is in debug mode, preserve default libav logs. This is to solve the `no frame!` spam.
1. Upgrade gortsplib from `v3.7.0` to `v4.8.0` (this was a pre req for getting rtp_passthrough to work)
1. Upgrade goutils from `v0.1.60` to `v0.1.71`
1. Upgrade rtp from `1.8.2` to `1.8.5`
1. Upgrade from go `1.19` to `1.21`
1. Add tests for rtp_passthrough, GetImage rtppasstrough.Source & config validations
1. Update Makefile to support building concurrently for multiple platforms

Test with viam-sever `5a86a87675b8276b7d4aa6699e474ec4b368a44a` (tip of main on 04/12/2024)

### Bug List
- [ ] Nick S: switching hardware camera to -> from h264 -> h265 sometimes results in the image not updating anymore in the GetImage code path, might be related to unsyncronized calls to libav not sure
- [ ] Nick S: switching hardware camera to -> from h264 -> h265 sometimes results in reconnect loop with non nil error who's string representation is empty string 
- [ ] Nick S: `\_ [hevc @ 0x13040e170] Could not find ref with POC 12` error log, non fatal, when using hardware h265 camera
- [ ] Nick S: `\_ [swscaler @ 0x310de8000] No accelerated colorspace conversion found from yuv420p to rgba.`  error log, non fatal, when using hardware h265 camera

### Tested Features:
1. `erh:viamrtsp:rtsp`
    1. H264 GetImage - physical camera 
    3. H264 Reconfigure - MediaMTX
    4. Passthrough streaming & GetImage - physical camera
    5. Passthrough Reconfigure
    6. H265 GetImage - physical camera 
    7. H265 Reconfigure - MediaMTX
    8. MJPEG GetImage - MediaMTX
    9. MJPEG Reconfigure - Media MTX
1.  `erh:viamrtsp:rtsp-h264`
    1.  H264 GetImage - physical camera 
    2. H264 Reconfigure - MediaMTX
    3. Passthrough
    4. Passthrough Reconfigure
    5. H265 - fail
    6. MJPEG - fail
1. `erh:viamrtsp:rtsp-h265`
    1. H264 GetImage - fail
    3. H264 Passthrough - fail
    5. H265 GetImage - physical camera 
    6. H265 Reconfigure - MediaMTX
    8. MJPEG GetImage - fail
    10. MJPEG Reconfigure - fail
1. `erh:viamrtsp:rtsp-mjpeg`
    1. H264 GetImage - fail
    3. Passthrough - fail
    5. H265 GetImage - fail
    7. MJPEG GetImage - MediaMTX
    8. MJPEG Reconfigure - Media MTX

### Tested Platforms:
- [ ] PI 4 (bullseye)
- [ ] Pi 5 ( bookworm)
- [ ] x86_64 linux (ubuntu 22.04.4)
- [ ] MacOS aarm64


cmd/module/cmd.go
```diff
@@ -8,24 +8,20 @@ import (
...
 	"github.com/erh/viamrtsp"
+	"go.viam.com/utils"
 )
 
 func main() {
-	err := realMain()
-	if err != nil {
-		panic(err)
-	}
+	utils.ContextualMain(mainWithArgs, module.NewLoggerFromArgs("client"))
```
nicksanford, line 15: I changed this as the current implementation doesn't call `Close` on itself when the process receives certain signals which resulted in the RTSP server the module was connected to having dangling connections after the module died or rebooted. 

[ContextualMain](https://github.com/viamrobotics/goutils/blob/main/runtime.go#L39) will catch these signals & will close the context & allow the module to call `Close()` on itself & all models (which can then release RTSP connections).

cmd/module/cmd.go
```diff
@@ -3,29 +3,29 @@ package main
...
-	logger := logging.NewDebugLogger("client")
 
+func mainWithArgs(ctx context.Context, args []string, logger logging.Logger) error {
 	myMod, err := module.NewModuleFromArgs(ctx, logger)
 	if err != nil {
 		return err
 	}
 
+	if logger.Level() != zapcore.DebugLevel {
+		viamrtsp.SetLibAVLogLevelFatal()
```
nicksanford, line 26: This gets rid of the spammy `no frame!` logs which don't impact video quality unless the module is configured to run in debug mode.

go.mod
```diff
@@ -162,3 +164,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	nhooyr.io/websocket v1.8.7 // indirect
 )
+
+replace go.viam.com/rdk => /Users/nicksanford/rdk
```
nicksanford, line 168: update after merge

.canon.yaml
```diff
@@ -1,14 +1,15 @@
-rtsp:
+viam-rtsp:
```
nicksanford, line 1: Updated to follow the pattern in RDK .canon.yaml & to update to use images that have go 1.21

cmd/module/cmd.go
nicksanford: If we merge in https://github.com/erh/viamrtsp/pull/5 before this change, this can be removed

rtsp_test.go
```diff
@@ -0,0 +1,383 @@
+package viamrtsp
```
nicksanford, line 1: Once this PR is merged, I'll create a ticket to run tests as part of CI

go.mod
```diff
@@ -162,3 +164,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	nhooyr.io/websocket v1.8.7 // indirect
 )
+
+replace go.viam.com/rdk => github.com/viamrobotics/rdk v0.24.0-rc0.0.20240405205559-fa9ce0d7f3bf
```
nicksanford, line 168: This is needed to ensure we have at least this change from RDK: https://github.com/viamrobotics/rdk/pull/3773 so we can do rtp_passthrough

rtsp.go
```diff
@@ -86,13 +99,20 @@ type rtspCamera struct {
...
 }
 
 // Close closes the camera. It always returns nil, but because of Close() interface, it needs to return an error.
 func (rc *rtspCamera) Close(ctx context.Context) error {
 	rc.cancelFunc()
+	rc.unsubscribeAll()
+	rc.closeConnection()
 	rc.activeBackgroundWorkers.Wait()
-	return rc.closeConnection()
+	return nil
```
nicksanford, line 115: We should first close all connections & then wait for all background goroutines to terminate, not vice versa

rtsp.go
```diff
@@ -130,17 +150,15 @@ func (rc *rtspCamera) clientReconnectBackgroundWorker(codecInfo videoCodec) {
 	}, rc.activeBackgroundWorkers.Done)
 }
 
-func (rc *rtspCamera) closeConnection() error {
-	var err error
+func (rc *rtspCamera) closeConnection() {
 	if rc.client != nil {
-		err = rc.client.Close()
+		rc.client.Close()
```
nicksanford, line 155: new version of the lib no longer returns an error

rtsp.go
```diff
@@ -151,13 +169,23 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
...
-	err = rc.closeConnection()
-	if err != nil {
-		rc.logger.Debugw("error while closing rtsp client:", "error", err)
-	}
+	rc.closeConnection()
 
 	// replace the client with a new one, but close it if setup is not successful
 	rc.client = &gortsplib.Client{}
+	rc.client.OnPacketLost = func(err error) {
+		rc.logger.Debugf("OnPacketLost: err: %s", err.Error())
```
nicksanford, line 177: Adds more observability in debug mode with RTSP faults occur

rtsp.go
```diff
@@ -166,12 +194,7 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
 		}
 	}()
 
-	err = rc.client.Start(rc.u.Scheme, rc.u.Host)
-	if err != nil {
-		return err
-	}
-
-	tracks, baseURL, _, err := rc.client.Describe(rc.u)
+	session, _, err := rc.client.Describe(rc.u)
```
nicksanford, line 197: New lib changed the signature

rtsp.go
```diff
@@ -207,26 +230,23 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
...
+	if media == nil {
 		rc.logger.Warn("tracks available")
-		for _, x := range tracks {
+		for _, x := range session.Medias {
 			rc.logger.Warnf("\t %v", x)
 		}
 		return errors.New("h264 track not found")
 	}
-
-	_, err = rc.client.Setup(track, baseURL, 0, 0)
```
nicksanford, line 223: Setup is now called here: https://github.com/erh/viamrtsp/pull/7/files#diff-9e6c919d1dd8bdada885f74ffe36d4f38ace393b0ef5dce7e30cedba4340fa7dR350

rtsp.go
```diff
@@ -207,26 +230,23 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
...
-	if err != nil {
-		return err
-	}
+	forma = f
 
 	// setup RTP/H264 -> H264 decoder
-	rtpDec, err := format.CreateDecoder2()
+	rtpDec, err := f.CreateDecoder()
```
nicksanford, line 249: New lib version

rtsp.go
```diff
@@ -239,71 +259,126 @@ func (rc *rtspCamera) initH264(tracks media.Medias, baseURL *url.URL) (err error
...
 			if err != rtph264.ErrNonStartingPacketAndNoPrevious && err != rtph264.ErrMorePacketsNeeded {
-				rc.logger.Errorf("error decoding(1) h264 rstp stream %v", err)
+				rc.logger.Debugf("error decoding(1) h264 rstp stream %v", err)
 			}
 			return
 		}
 
-		for _, nalu := range au {
+		if !iFrameReceived {
+			if !h264.IDRPresent(au) {
```
nicksanford, line 285: This solves most of the `no frame!` errors

rtsp.go
```diff
@@ -239,71 +259,126 @@ func (rc *rtspCamera) initH264(tracks media.Medias, baseURL *url.URL) (err error
...
+			}
+			if image != nil {
+				rc.latestFrame.Store(&image)
 			}
+		}
+	}
 
-			// convert NALUs into RGBA frames
-			lastImage, err := rc.rawDecoder.decode(nalu)
+	onPacketRTP := func(pkt *rtp.Packet) {
```
nicksanford, line 308: If passthrough is not enabled we only want to execute the `storeImage` callback to maintain current behavior

rtsp.go
```diff
@@ -239,71 +259,126 @@ func (rc *rtspCamera) initH264(tracks media.Medias, baseURL *url.URL) (err error
...
+			for _, subAndCB := range rc.subAndCBByID {
+				if err := subAndCB.sub.Publish(func() error { return subAndCB.cb(u) }); err != nil {
+					rc.logger.Debug("RTP packet dropped due to %s", err.Error())
+				}
 			}
 		}
-	})
+
+		onPacketRTP = func(pkt *rtp.Packet) {
+			publishToWebRTC(pkt)
```
nicksanford, line 345: if passthrough is enabled, we want to both publish to WebRTC & storeImage

go.mod
```diff
@@ -1,13 +1,19 @@
 module github.com/erh/viamrtsp
 
-go 1.19
+go 1.21
```
nicksanford, line 3: upgrade to go 1.21 as min version

go.mod
```diff
@@ -1,13 +1,19 @@
...
+
+toolchain go1.21.8
 
 require (
-	github.com/bluenviron/gortsplib/v3 v3.7.0
-	github.com/pion/rtp v1.8.2
+	github.com/bluenviron/gortsplib/v4 v4.8.0
```
nicksanford, line 8: upgrade to gortsplib/v4

go.mod
```diff
@@ -162,3 +164,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	nhooyr.io/websocket v1.8.7 // indirect
 )
+
+replace go.viam.com/rdk => github.com/viamrobotics/rdk v0.24.0-rc0.0.20240410143850-e4e67584717a
```
nicksanford, line 168: Upgrade to this commit of RDK: https://github.com/viamrobotics/rdk/pull/3788

Makefile
```diff
@@ -55,3 +55,6 @@ module: bin/viamrtsp
 
 clean:
 	rm -rf FFmpeg bin/viamrtsp module.tar.gz
+
+clean-all:
```
seanavery, line 59: should we combine this with clean?

go.mod
seanavery: Do we want to keep this replace statement or just update the require section?

rtsp.go
```diff
@@ -38,6 +42,7 @@ var ModelH264 = family.WithModel("rtsp-h264")
 var ModelH265 = family.WithModel("rtsp-h265")
 var ModelMJPEG = family.WithModel("rtsp-mjpeg")
 var Models = []resource.Model{ModelAgnostic, ModelH264, ModelH265, ModelMJPEG}
+var ErrH264PassthroughNotEnabled = errors.New("H264 passthrough is not enabled")
```
seanavery, line 45: [nit] Not a gorista, but should we format in a var ( ) block here?

```
var (
    ModelH265                   = family.WithModel("rtsp-h265")
    ModelMJPEG                  = family.WithModel("rtsp-mjpeg")
    Models                      = []resource.Model{ModelAgnostic, ModelH264, ModelH265, ModelMJPEG}
    ErrH264PassthroughNotEnabled = errors.New("H264 passthrough is not enabled")
)
```

rtsp.go
```diff
@@ -213,25 +235,23 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec) (err error) {
...
 			rc.logger.Warnf("\t %v", x)
 		}
 		return errors.New("h264 track not found")
 	}
-
-	_, err = rc.client.Setup(track, baseURL, 0, 0)
-	if err != nil {
-		return err
-	}
+	forma = f
```
seanavery, line 251: why are we doing this instead of just using f?

rtsp.go
```diff
@@ -399,7 +470,95 @@ func (rc *rtspCamera) initMJPEG(tracks media.Medias, baseURL *url.URL) error {
...
+
+		// If we have no AUs we can't encode packets.
+		if tunit.AU == nil {
+			return nil
+		}
+
+		if !firstReceived {
+			firstReceived = true
+		} else if tunit.PTS < lastPTS {
+			return errors.New("WebRTC doesn't support H264 streams with B-frames")
```
seanavery, line 522: What is expected to happen if h264 feed uses b frames?

rtsp.go
```diff
@@ -399,7 +470,95 @@ func (rc *rtspCamera) initMJPEG(tracks media.Medias, baseURL *url.URL) error {
 
 		rc.latestFrame.Store(&img)
 	})
+	return nil
+}
+
+// SubscribeRTP registers the PacketCallback which will be called when there are new packets.
+// NOTE: Packets may be dropped before calling packetsCB if the rate new packets are received by
+// the rtppassthrough.Source is greater than the rate the subscriber consumes them.
```
seanavery, line 478: Have we been able to create this condition? Can the decoder on the webapp side recover from missing packets once passed through?

rtsp.go
seanavery: Curious to try this out with the input camera using different libx264 b frame strategies. Such as:

```
-vcodec libx264 -pix_fmt yuv420p -bf 3 -b_strategy 2 
``

File: Makefile (hunk 19 above)
Comment (nicksanford) [reply Makefile:60]: I think `clean-all` should be separate as it will delete anything that is not in version control, so it could easily lead to loss of work if done mistakenly. I think `clean` should clean the things that we expect people to want to clean frequently.

I've also updated the makefile to separate out build directories for each version of ffmpeg (currently only n6.1) and platform so that we should never need to clean the FFmpeg build directory when we switch platforms.

File: go.mod (hunk 18 above)
Comment (nicksanford) [reply go.mod:168]: I would be in favor of keeping the replace for now and remove the replace when the passthrough features land in a tagged release of RDK (which should be next tuesday)

File: rtsp.go (hunk 20 above)
Comment (nicksanford) [reply rtsp.go]: good call, done

File: rtsp.go (hunk 22 above)
Comment (nicksanford) [reply rtsp.go:524]: This is inspired by https://github.com/bluenviron/mediamtx/blob/main/internal/servers/webrtc/session.go#L185 (I've commented the code as such)

Right now it will just kill the video stream subscription & the caller won't get packets if the H264 stream has B-frames. 

In the future, when I get https://github.com/viamrobotics/rdk/pull/3792 working, I intend to change the signature of camera.SubscribeRTP such that the caller can detect when the subscription dies (such as happens in this case) & act accordingly, by, for example, falling back to GetImage.

File: Makefile (hunk 19 above)
Comment (seanavery) [reply Makefile:60]: Sounds good.

File: rtsp.go (hunk 21 above)
Comment (nicksanford) [reply rtsp.go]: good call, removed

File: Makefile [hunk 24]
```diff
@@ -14,44 +18,45 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
...
 	go mod tidy
 
-FFmpeg:
-	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch n6.1
+$(FFMPEG_VERSION_PLATFORM):
+	git clone https://github.com/FFmpeg/FFmpeg.git --depth 1 --branch $(FFMPEG_TAG) $(FFMPEG_VERSION_PLATFORM)
 
-$(FFMPEG_PREFIX): FFmpeg
-	cd FFmpeg && ./configure $(FFMPEG_OPTS) && $(MAKE) -j$(shell nproc) && $(MAKE) install
+$(FFMPEG_BUILD): $(FFMPEG_VERSION_PLATFORM)
```
nicksanford, line 43: This will now have separate build directories per ffmpeg version & platform. I did this so that I can `make` from both `canon -arch amd64`, `canon -arch arm64` and from the OSX host & they can all safely build in parallel.

This dramatically sped up my ability to test different platforms

nicksanford:
```bash
ls FFmpeg/n6.1/
Darwin-arm64  Linux-aarch64 Linux-x86_64
```

Makefile
```diff
@@ -14,44 +18,45 @@ FFMPEG_OPTS ?= --prefix=$(FFMPEG_PREFIX) \
...
+.PHONY: build-ffmpeg lint update-rdk module clean clean-all
 
-bin/viamrtsp: build-ffmpeg *.go cmd/module/*.go
-	PKG_CONFIG_PATH=$(FFMPEG_PREFIX)/lib/pkgconfig \
+$(BIN_OUTPUT_PATH)/viamrtsp: build-ffmpeg *.go cmd/module/*.go
+	PKG_CONFIG_PATH=$(FFMPEG_BUILD)/lib/pkgconfig \
 		CGO_LDFLAGS=$(CGO_LDFLAGS) \
-		go build -o bin/viamrtsp cmd/module/cmd.go
-
-test:
```
nicksanford, line 29: `make test` didn't work (I think b/c the `CGO_LDFLAGS` cause it to try to dynamically load the libav libs, but `go test` outside of make still works. 

As a result, I removed the target that didn't work

rtsp.go
```diff
@@ -244,21 +264,19 @@ func (rc *rtspCamera) initH264(tracks media.Medias, baseURL *url.URL) (err error
 	}
 
 	// if SPS and PPS are present into the SDP, send them to the decoder
-	if format.SPS != nil {
-		rc.rawDecoder.decode(format.SPS)
+	if f.SPS != nil {
+		rc.rawDecoder.decode(f.SPS) // nolint:errcheck
```
randhid, line 269: Is there a reason not to add the error checks?

rtsp.go
```diff
@@ -315,28 +383,28 @@ func (rc *rtspCamera) initH265(tracks media.Medias, baseURL *url.URL) (err error
 	}
 
 	// For H.265, handle VPS, SPS, and PPS
-	if format.VPS != nil {
-		rc.rawDecoder.decode(format.VPS)
+	if f.VPS != nil {
+		rc.rawDecoder.decode(f.VPS) // nolint:errcheck
```
randhid, line 380: same here re: error checks, also okay to disregard if this was already the behaviour and one of us will do it in a separate ticket.

rtsp.go
```diff
@@ -445,6 +609,15 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 	return camera.FromVideoSource(conf.ResourceName(), src, logger), nil
 }
 
+func (rc *rtspCamera) unsubscribeAll() {
+	rc.subsMu.Lock()
+	defer rc.subsMu.Unlock()
+	for id, subAndCB := range rc.subAndCBByID {
+		subAndCB.sub.Close()
```
randhid, line 621: could this ever error, and if so what are the consequences?

seanavery (commented):
Feedback from testing out agnostic `rtsp` model.

If `rtp_passthrough` is enabled, and the endpoint does not support h264 but does have tracks available for another supported codec (h265, mjpeg), I would expect as a user that the module would fallback to get image things would still work.

Instead we error out saying:
```
resource/graph_node.go:230  resource build error: rpc error: code = Unknown desc = address reports to have only an h265 track but rtpH264Passthrough was enabled  {"resource":"rdk:component:camera/ip-cam","model":"erh:viamrtsp:rtsp"}
```

nicksanford:
This PR has been split up into:
Build improvements: https://github.com/erh/viamrtsp/pull/16
Deps update: https://github.com/erh/viamrtsp/pull/17
The actual feature: https://github.com/erh/viamrtsp/pull/19

PR #8, opened 2024-03-28
[RSDK-6821] Add integration test workflow

hexbabe:
https://viam.atlassian.net/browse/RSDK-6821

This CI [works on my fork](https://github.com/hexbabe/sean-viamrtsp/pull/1). Once we enable CI runners to pick the workflow up and verify it works upstream here, this will be ready for final review.

Edit: [it now also works upstream](https://github.com/erh/viamrtsp/actions/runs/8529726297/job/23366098458).


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,78 @@
+name: Integration test with mediamtx stream
+
+on: [push]
+
+jobs:
+  build-and-test:
```
seanavery, line 6: Might be a good idea to set up a `matrix` here to run build/test on both x86 and arm64.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.19
+    
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo add-apt-repository universe &&
+        sudo apt-get install -y libswscale-dev libavcodec-dev libavformat-dev libavutil-dev ffmpeg libfuse2
```
seanavery, line 21: Pretty sure we can remove the *-dev libs and just apt install ffmpeg for fake rtsp cam. (Once static build PR is in). But with the current build/runtime we will indeed need these.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Install viam-server
+      run: |
+        curl https://storage.googleapis.com/packages.viam.com/apps/viam-server/viam-server-stable-x86_64.AppImage -o viam-server
+        chmod 755 viam-server
+        sudo ./viam-server --aix-install
+
+    # Generate config from within yml because relative paths don't reliably work for module exec paths
+    - name: Generate viam-server config
+      run: |
+        echo '{
```
seanavery, line 46: I like this! We might want to setup templates in a big json file though to test out the different models.
The run cmd could use `jq` to parse and then cout to file like you are doing.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,78 @@
...
+              "executable_path": "'$(realpath bin/viamrtsp)'"
+            }
+          ]
+        }' > integration-test-config.json
+        
+    # Sleep to give time for server to start
+    - name: Run viam-server
+      run: |
+        viam-server -debug -config ./integration-test-config.json &
+        sleep 10
```
seanavery, line 72: Is this timeout reliable?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,78 @@
...
+                "rtsp_address": "rtsp://localhost:8554/live.stream"
+              },
+              "depends_on": []
+            }
+          ],
+          "modules": [
+            {
+              "type": "local",
+              "name": "viamrtsp",
+              "executable_path": "'$(realpath bin/viamrtsp)'"
```
seanavery, line 63: Nice!

seanavery (commented):
Looking good! 
- Test script for client is simple and seems like it does the job.
    - Do not think we need to use the testing package.
- Added some comments about extending tests to cover model names and architectures.
    - (Realize model change has not been merged yet..)

.github/workflows/integration-test.yml
hexbabe: Got it will keep that back of mind

.github/workflows/integration-test.yml
hexbabe: I'm not sure if it's necessary. Let me try it out without

.github/workflows/integration-test.yml
hexbabe: We should chat about how we are going to handle that

> big json file

Would we pre-write these files and then populate the module path programmatically in the .yml script?

.github/workflows/integration-test.yml
hexbabe: I don't think arm64 runners are working properly... https://github.com/hexbabe/sean-viamrtsp/actions/runs/8484610925/job/23247871489?pr=1 <- this run should fail because I forgot to update the binaries (viam-server and mediamtx) to be for `arm64`, yet the CI still passed. I think in viam-robotics we use buildjet to do arm64 CI, we don't have that here

.github/workflows/integration-test.yml
hexbabe: Haha yeah the new wait-for-port-to-be-available logic is way better than the sleep. Thanks for the callout!

seanavery: Ok as long as this is not flaky, I am ok with it.

Here is a bash script i wrote to wait for viam-server if helpful.
(waits for the web server to boot up)
```
elapsed_time=0
echo "waiting for localhost:8080..."
while ! nc -z localhost 8080; do
  sleep "$WAIT_INTERVAL"
  if [ "$elapsed_time" -ge "$MAX_WAIT_TIME" ]; then
    echo "Timed out waiting for localhost:8080"
    exit 1
  fi
  elapsed_time=$((elapsed_time + WAIT_INTERVAL))
done
```

randhid (approved):
trying to approve to see if it picks up your workflow

hexbabe:
@erh this is also ready for ur final review + merge

erh (approved):

PR #9, opened 2024-03-29
[RSDK-6937] Add MJPEG codec

hexbabe:
https://viam.atlassian.net/browse/RSDK-6937


rtsp.go
```diff
@@ -198,9 +203,48 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
+
+	mjpegDecoder, err := mjpegFormat.CreateDecoder2()
+	if err != nil {
+		return errors.Wrap(err, "error creating MJPEG decoder")
+	}
+
+	rc.client.OnPacketRTP(track, mjpegFormat, func(pkt *rtp.Packet) {
+		frame, _, err := mjpegDecoder.Decode(pkt)
+		if err != nil {
+			return
```
hexbabe, line 231: I'm not sure why there are so many garbage frames that we just ignore the error here. I am mirroring https://github.com/viamrobotics/rdk/blob/0a7f6b23851f8c04f82ffa5583f600668e5e2e3a/components/camera/rtsp/rtsp.go#L171

I also checked and even if we ignore the error here, we are still getting a full 60fps stream on chrome webrtc internals

rtsp.go
```diff
@@ -352,6 +356,49 @@ func (rc *rtspCamera) initH265(tracks media.Medias, baseURL *url.URL) (err error
...
```
hexbabe, line 384: I'm not sure why there are so many garbage frames that we just ignore the error here. I am mirroring https://github.com/viamrobotics/rdk/blob/0a7f6b23851f8c04f82ffa5583f600668e5e2e3a/components/camera/rtsp/rtsp.go#L171

I also checked and even if we ignore the error here, we are still getting a full 60fps stream on chrome webrtc internals

decoder.go
```diff
@@ -37,18 +37,17 @@ int get_video_codec(AVFormatContext *avFormatCtx) {
         if (codecParams->codec_id == AV_CODEC_ID_H264) {
             return AV_CODEC_ID_H264;
         } else if (codecParams->codec_id == AV_CODEC_ID_H265) {
-            found_h265 = 1;
+            return AV_CODEC_ID_H265;
```
seanavery, line 40: This may not be prioritized correctly because you could in theory have multiple tracks for a given endpoint in any order. So if returning inside the track iterator here, you may never see the higher priority codec. In practice, I doubt this would come up very often as you normally would only have one video codec for an endpoint.

rtsp.go
seanavery: Do we know what errors we are seeing? Should we debug log here?

rtsp.go
```diff
@@ -352,6 +356,49 @@ func (rc *rtspCamera) initH265(tracks media.Medias, baseURL *url.URL) (err error
...
+		}
+		if frame == nil {
+			return
+		}
+
+		img, err := jpeg.Decode(bytes.NewReader(frame))
```
seanavery, line 390: Make sense to me to simply use image/jpeg instead of FFmpeg decode here.

rtsp.go
hexbabe: the error is "need more packets"

hexbabe: It's suuuper duper spammy. Might even drown out other debug logs

rtsp.go
hexbabe: Yeah should just be JPEG frames we are dealing with

nicksanford:
Now that we have end to end tests could we please add an end to end test for Mjpeg?

decoder.go
hexbabe: Discussed offline: we will leave the incorrect prioritization in this PR where @seanavery will pick the logic up and refactor it

hexbabe:
> Now that we have end to end tests could we please add an end to end test for Mjpeg?

Done @nicksanford

seanavery (approved):
LGTM

hexbabe:
@nicksanford this is ready for re-review... would directly add you as reviewer, but I think Eliot needs to add you as a maintainer

randhid (approved):

seanavery (commented):
- Do we want to add a model for this`rtsp-mjpeg`?

hexbabe:
> * Do we want to add a model for this`rtsp-mjpeg`?

My thought process was that we would want to implicitly test the codec discovery logic using the agnostic model for all streams

hexbabe:
Tested manually on Darwin arm64 https://www.loom.com/share/42411fade50a42f7b7a63aeb8f225d31?sid=4f8cbbba-876e-4d25-ae02-7f2df1882468

nicksanford:
@hexbabe looks good, I'm good with you merging.
Def create a ticket for https://github.com/erh/viamrtsp/pull/9/files#r1550407412 to be sure it doesn't get lost in the shuffle.

hexbabe:
> @hexbabe looks good, I'm good with you merging. Def create a ticket for https://github.com/erh/viamrtsp/pull/9/files#r1550407412 to be sure it doesn't get lost in the shuffle.

Yes yes good call. @seanavery is #13 the aforementioned ticket?

PR #11, opened 2024-04-05
bump-go-version-to-1.21

nicksanford:
https://github.com/erh/viamrtsp/pull/7 depends on go 1.21 (which is the currently supported go version in RDK) 

This updates CI to use go 1.21 so that that PR can pass tests


hexbabe (approved):
Makes sense

seanavery (approved):

PR #12, opened 2024-04-08
RSDK-7226 Add chunks flag to decoder

seanavery:
## Description

The goal of this PR is to remove the 20 byte filter from h264 decoder loop by properly flagging the decoder to expect partial frames. This is needed since we are feeding NALs into the decoder instead of by frame boundary.

## Testing
- Tested that the decoder no longer logs `no frame!`.
- No regression to h264/h265 decoding performance or stability.

## Refs
- [AV_CODEC_FLAG2_CHUNKS](https://ffmpeg.org/doxygen/2.8/group__lavc__core.html#gabc5592664ff0686def0c2b41cfcc322d)

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -133,6 +133,9 @@ func newDecoder(codecID C.enum_AVCodecID) (*decoder, error) {
 		return nil, fmt.Errorf("avcodec_alloc_context3() failed")
 	}
 
+	// Set the decoder to handle partial frames
+	codecCtx.flags2 |= C.AV_CODEC_FLAG2_CHUNKS
+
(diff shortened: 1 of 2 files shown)


seanavery:
@nicksanford Tagging for review.

nicksanford:
@seanavery Great find! This looks good to me. I tested it on arm64 linux (pi 4 bullseye) and macos aarm64. If you have tested on pi 5 bookworm & amd64 linux this is good to merge from my perspective.

seanavery:
- Amd64 ✅ 
- Pi5 (bworm) ✅

PR #13, opened 2024-04-10
RSDK-7196 Codec lookup without libav

seanavery:
## Description

Currently, when using `Agnostic` model `rtsp`, libav is used to inspect the streams and parse out codecs. Instead, we can use the SDP parsers in gortsplib to get that information. This will reduce C dependencies and speed up boot time for the module.

## Testing
Testing using the `Agnostic` model to trigger codec lookups.

-  erh:viamrtsp:rtsp (Get Image)
    - codec lookup works for h264, h265, mjpeg ✅ 
    - get image works for h264, h264, mjpeg ✅ 
    - change streams from h264->h265, h265->h264 confirm recovers ✅ 
    - Reconfigure ✅ 
    - Reconfigure by changing URL works ✅ 

 - erh:viamrtsp:rtsp (rtp_passthrough)
   - passthrough working with h264 endpoint ✅ 
       - warns when not h264 `rtp_passthrough is only supported for H264 codec. rtp_passthrough features disabled due to H265 RTSP track` ✅ 
    - passthrough and GetImage change streams h264->h265, h265->h264 confirm recovers ✅ 
      - Does not go pack to passthrough but GetImage still working (as expected)
    - Reconfigure ✅ 
    - Reconfigure by changing URL works ✅


decoder.go
```diff
@@ -74,42 +45,23 @@ func frameLineSize(frame *C.AVFrame) *C.int {
...
-	}
-	defer C.avformat_close_input(&avFormatCtx)
-
-	ret = C.avformat_find_stream_info(avFormatCtx, nil)
-	if ret < 0 {
-		return Unknown, fmt.Errorf("avformat_find_stream_info() failed: %s", avError(ret))
-	}
+	// List of formats/codecs in priority order
+	formatPointers := []interface{}{&h264, &h265}
+	codecs := []videoCodec{H264, H265}
```
seanavery, line 55: `formatPointers` and `codecs` list here could be combined into a struct if that is cleaner.

```
    type formatAndCodec struct {
        formatPtr interface{}
        codec     videoCodec
    }
 ```

File: decoder.go [hunk 2]
```diff
@@ -74,42 +45,23 @@ func frameLineSize(frame *C.AVFrame) *C.int {
 	return (*C.int)(unsafe.Pointer(&frame.linesize[0]))
 }
 
-// getStreamInfo opens a stream URL and retrieves the video codec.
-func getStreamInfo(url string) (videoCodec, error) {
-	cUrl := C.CString(url)
-	defer C.free(unsafe.Pointer(cUrl))
+// getAvailableCodec returns a supported codec in the given tracks.
+func getAvailableCodec(tracks media.Medias) videoCodec {
```
nicksanford, line 49: Could we please move any functions that don't have any c deps out of this file?

decoder.go
```diff
@@ -74,42 +45,23 @@ func frameLineSize(frame *C.AVFrame) *C.int {
...
-		return Unknown, fmt.Errorf("avformat_open_input() failed: %s", avError(ret))
-	}
-	defer C.avformat_close_input(&avFormatCtx)
-
-	ret = C.avformat_find_stream_info(avFormatCtx, nil)
-	if ret < 0 {
-		return Unknown, fmt.Errorf("avformat_find_stream_info() failed: %s", avError(ret))
-	}
+	// List of formats/codecs in priority order
+	formatPointers := []interface{}{&h264, &h265}
```
nicksanford, line 54: this should be a slice of `format.Format` not interface{} see https://github.com/erh/viamrtsp/pull/7/files#diff-616e82b0c6eddaf46a49b2e6fa5146effd4c74559a0f0f152df3d330d6f4d9e4R35

decoder.go
nicksanford: yes that is cleaner IMO

nicksanford:
I like the direction this PR is going. Will re-review with the feedback is addressed & this PR has been rebased on main after the MJPEG support has landed.

decoder.go
seanavery: Yes, I was trying to use `format.Format` here but was running into issues.

- The library shows using a pointer to the format and then memory ref which is essentially a double pointer...
    - The double pointer no longer satisfies the interface
```
./rtsp.go:421:37: cannot use &h264 (value of type **formats.H264) as type formats.Format in array or slice literal:
        **formats.H264 does not implement formats.Format (missing ClockRate method)
./rtsp.go:421:44: cannot use &h265 (value of type **formats.H265) as type formats.Format in array or slice literal:
        **formats.H265 does not implement formats.Format (missing ClockRate method)
make: *** [Makefile:24: bin/viamrtsp] Error 2
```
- If I just use a single pointer, the interface is satisfied. However, the `findFormat` fails (does not pick up any formats).
- I also tried creating new format instances using `new` without success. (no format is found).

seanavery: Ah, I see you are upgraded to `v4` in that PR. https://github.com/erh/viamrtsp/pull/7/files#diff-616e82b0c6eddaf46a49b2e6fa5146effd4c74559a0f0f152df3d330d6f4d9e4R10

hexbabe:
https://github.com/erh/viamrtsp/pull/14 is in main now! Can rebase

decoder.go
seanavery: I have not been able to successfully use `format.Format`.

`FindFormat` takes in an empty interface:
```
// FindFormat finds a certain format among all the formats in all the medias of the stream.
// If the format is found, it is inserted into forma, and its media is returned.
func (d *Session) FindFormat(forma interface{}) *Media {
	for _, media := range d.Medias {
		ok := media.FindFormat(forma)
		if ok {
			return media
		}
	}
	return nil
}
```

This aligns with demo code that passes in the memory address of the pointer to the format (which does not satisfy the Format interface).

randhid (approved):

rtsp.go
```diff
@@ -475,3 +474,26 @@ func modelToCodec(model resource.Model) (videoCodec, error) {
 		return Unknown, fmt.Errorf("model '%s' has unspecified codec handling", model.Name)
 	}
 }
+
+// getAvailableCodec determines the first supported codec from a session's SDP data
+// returning Unknown if none are found.
+func getAvailableCodec(tracks *description.Session) videoCodec {
```
nicksanford, line 480:
```suggestion
func getAvailableCodec(session *description.Session) videoCodec {
```

rtsp.go
```diff
@@ -53,6 +53,12 @@ type Config struct {
 	DistortionParams *transform.BrownConrady            `json:"distortion_parameters,omitempty"`
 }
 
+// CodecFormat contains a pointer to a format and the corresponding FFmpeg codec.
+type CodecFormat struct {
```
nicksanford, line 57: lets make this type & its members private please as we don't expect them to be used outside of the viamrtsp package.

nicksanford (commented):
Can you please rebase on main & list all manual testing you have done. 

We should do manual testing similar to the checklist lisetd in ttps://github.com/erh/viamrtsp/pull/19

seanavery:
@nicksanford Rebased on main -- will report testing in description.

decoder.go
```diff
@@ -4,46 +4,15 @@ import (
 	"fmt"
 	"image"
 	"unsafe"
-
-	"github.com/pkg/errors"
 )
 
 /*
 #cgo pkg-config: libavcodec libavutil libswscale libavformat
```
nicksanford, line 10: Lets remove `libavformat` from the `pkg-config` as we no longer depend on it

nicksanford (approved):
Just one last suggestion. Otherwise looks great! The camera boots much faster which is a big UX improvement when you have a lot of cameras.

PR #22, opened 2024-04-23
revert-RSDK-7226

nicksanford:
Reverts https://github.com/erh/viamrtsp/pull/12/files due to it causing https://viam.atlassian.net/browse/RSDK-7346 / https://viam.atlassian.net/browse/RSDK-7354

This is a temp fix to revert back to `v0.0.1` behavior. @seanavery is working on a more long term fix


randhid (approved):

seanavery (approved):
LGTM.

Seems ok to revert to using a length check instead of an explicit SPS/PPS type filter.

PR #24, opened 2024-05-01
RSDK-7390

nicksanford:


rtsp.go
```diff
@@ -287,21 +290,13 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
...
-				rc.logger.Errorf("error decoding(2) h264 rtsp stream  %s", err.Error())
-				return
-			}
-			if image != nil {
-				rc.latestFrame.Store(&image)
-			}
+		if !receivedFirstIDR && h264.IDRPresent(au) {
+			rc.logger.Debug("adding inital SPS & PPS")
+			receivedFirstIDR = true
+			au = append(initialSPSAndPPS, au...)
```
nicksanford, line 296: I confirmed that if the SPS & PPS are also sent in the first IDR au (so [SPS, PPS, SPS, PPS, IDR]) that is able to be decoded into a jpeg without causing any error log spam.

seanavery: nice, that is handy! So it seems if the first IDR has different SPS and PPS from the SDP-derived ones, they will be overwritten by the ordering here.

rtsp.go
```diff
@@ -705,3 +697,74 @@ func getAvailableCodec(session *description.Session) videoCodec {
...
+func (rc *rtspCamera) compactH264SPSAndPPSAndIDR(au [][]byte) ([]byte, int) {
+	compactedNALU, numCompacted := []byte{}, 0
+	for _, nalu := range au {
+		if !isCompactableH264(nalu) {
+			// return once we hit a non SPS, PPS or IDR message
+			return compactedNALU, numCompacted
+		}
+		// If this is the first iteration, don't add the start code
+		// as the first nalu has not been written yet
+		if len(compactedNALU) > 0 {
```
seanavery, line 737: Makes sense for compatibility with `decode` that prepends start code.

rtsp.go
```diff
@@ -705,3 +697,74 @@ func getAvailableCodec(session *description.Session) videoCodec {
...
+			compactedNALU = append(compactedNALU, startCode...)
+		}
+		compactedNALU = append(compactedNALU, nalu...)
+		numCompacted++
+	}
+	return compactedNALU, numCompacted
+}
+
+// H2645StartCode is start code byte sequence for H264/H265 NALs.
+func H2645StartCode() []byte {
```
seanavery, line 748: possible [nit] - should we make this a const?

seanavery (approved):
LGTM!

mod offline discussion about mediamtx compatibility and possible FFmpeg version change.

rtsp.go
nicksanford: I don't think so given that append can mutate the underlying slice, which means that  there is a risk that data from other messages could leak into the global

PR #26, opened 2024-05-06
small-refactor: remove verbose logging & cancel cause

nicksanford:
Changes:
1. Remove the verbose `SubscribeRTP|Unsubscribe Start|Stop` log added during debugging
2. Change to cancelFn rather than cancelCuaseFn as the cause is not needed.


seanavery (approved):
LGTM.

PR #32, opened 2024-05-28
ignore tyy

hexbabe:


meta.json
```diff
@@ -21,5 +21,10 @@
       "model": "erh:viamrtsp:rtsp-mjpeg"
     }
   ],
-  "entrypoint": "bin/viamrtsp"
```
nicksanford, line 24: Can you confirm with someone on the registry team that changing the entrypoint between different module versions is safe (I think it is but just want to confirm)?

build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+sudo apt-get update
+sudo apt-get install -y libfuse2 ffmpeg pkg-config
```
nicksanford, line 5: 1. why do we need `libfuse2`?
2. why do we need ffmpeg if we are linking against it statically?

nicksanford (commented):
This is a super clean & minimal change. Good work! 

Just a few questions about whether we can trim up dependencies & to confirm that the `entrypoint` is allowed to change between module versions.

build.sh
hexbabe: Removed and verified the build succeeds and still works

meta.json
hexbabe: Confirmed it should be fine

build.sh
hexbabe: Removed "seanorg" stuff reverting back to erh/viamrtsp for everything. Should be good to merge if no more suggestions @nicksanford

nicksanford (approved):
This LGTM

I'll let you coordinate with Eliot on when he has added the required creds for the workflow to work before & trust you won't merge before that happens.

Makefile
```diff
@@ -145,11 +145,11 @@ endif
 endif
 endif
 
-module: $(BIN_OUTPUT_PATH)/viamrtsp
-	tar czf $(BIN_OUTPUT_PATH)/module.tar.gz $(BIN_OUTPUT_PATH)/viamrtsp
+module.tar.gz: $(BIN_OUTPUT_PATH)/viamrtsp
+	tar czf module.tar.gz -C $(BIN_OUTPUT_PATH) .
```
nicksanford, line 149: now that I've gone though the process of uploading a module I do have some more feedback here:
Can we please produce tarballs under the architecture that they are for:
```
dist/linux/amd64/module.tar.gz
dist/linux/arm64/module.tar.gz
dist/android/arm64/module.tar.gz
dist/darwin/arm64/module.tar.gz
```

We can also change the target back to just `module`

nicksanford: otherwise one could easily upload a module for one architecture to the registry under a different architecture. Splitting out the tarballs under a subdirectory for each platform reduces this risk.

meta.json
```diff
@@ -21,5 +21,10 @@
...
-}
\ No newline at end of file
+  "build": {
+    "build": "bash build.sh",
+    "path": "module.tar.gz",
+    "arch" : ["linux/amd64", "linux/arm64"]
+  },
+  "entrypoint": "viamrtsp"
```
nicksanford, line 29: I recommend not changing the entry point as eliot's fix to the module deploy problem didn't change the entrypoint: https://github.com/erh/viamrtsp/commit/4d7999f204a58b6564cfe2c0386e330fbf6c6f5c

Makefile
```diff
@@ -145,11 +145,11 @@ endif
 endif
 endif
 
-module: $(BIN_OUTPUT_PATH)/viamrtsp
-	tar czf $(BIN_OUTPUT_PATH)/module.tar.gz $(BIN_OUTPUT_PATH)/viamrtsp
+module.tar.gz: $(BIN_OUTPUT_PATH)/viamrtsp
```
nicksanford, line 148: please change the target back to `module` and change it to use this pattern: https://github.com/erh/viamrtsp/commit/4d7999f204a58b6564cfe2c0386e330fbf6c6f5c

build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+sudo apt-get update
+sudo apt-get install -y pkg-config
+
+make module.tar.gz
```
nicksanford, line 7: update to `make module`

hexbabe:
@nicksanford We can't easily use `BIN_OUTPUT_PATH` to specify the tarball path in `build.sh` or `meta.json`, so I added a new make target `module-deploy` that copies over the tarball to somewhere platform agnostic

nicksanford (approved):
LGTM

meta.json
```diff
@@ -1,25 +1,30 @@
 {
-  "module_id": "erh:viamrtsp",
-  "visibility": "public",
-  "url": "https://github.com/erh/viamrtsp",
+  "module_id": "seanorg:seanviamrtsp",
```
randhid, line 2: Does this need to change back at this point or still a todo? Since Nick just approved this pr.

hexbabe: Yep I will triple check before merging. I have it like this right now in case we need to make more test releases in my org. It's easy to revert I just apply this diff basically https://github.com/erh/viamrtsp/pull/32/commits/44df9a4a6e85d61ce505f5740d0cd0bf073a203f

randhid (approved):

erh (approved):

PR #37, opened 2024-06-12
RSDK-7808 - Add publish workflow for Android

seanavery:
## Description

Simple PR to integrate Android into the publish workflow.
- Need to use the older [viamrobotics/upload-module](https://github.com/viamrobotics/upload-module) action since we are cross-compiling for Android.
- [Here](https://github.com/seanavery/viamrtsp/actions/runs/9489033439/job/26149390554) is a successful workflow run on my forked repo pointing to private org.

## Tests 
- Tested module runtime from registry and works as expected from private org upload.


hexbabe:
Nice! Did you:
1. Test the upload action on in your own org in a private module
2. See that it uploaded to the registry and then 
3. Test the module from the registry on a robot

hexbabe:
Also product/release schedule question: Are we gonna deploy tag 0.0.2 to Android after this is merged in?

hexbabe (changes requested):
Nice! Honestly, I wonder if we can consolidate the two .ymls into this new one you wrote. May have to just add the linux targets and then add a conditional for `TARGET_OS` and `TARGET_ARCH`?

Can also just merge this and then make a follow-up ticket

seanavery:
@hexbabe The only issue i see with this combined workflow is that the `linux/arm64` upload seems to actually upload a `linux/amd64` binary... Perhaps I am looking at it wrong. But if that is the case, it might make more sense to combine once we get QEMU working.

.github/workflows/publish.yml
```diff
@@ -7,11 +7,50 @@ on:
...
+
+    - name: Install canon
+      run: |
+        go install github.com/viamrobotics/canon@latest
+    
+    - name: Launch canon
+      run: |
+        canon -profile viam-rtsp-antique -arch ${{ matrix.canon_arch }}
+
+    - name: Build module
```
hexbabe, line 44: I'm slightly suspicious that this might be running not in canon. Did you get around to verifying that all these executables run on their target archs?

seanavery: Good call! That would make sense - looking into it.

seanavery: Yep... arm64 upload is indeed x86_64 binary.

```
file /home/viam/.viam/packages/data/module/eacb3d64-4eab-43db-a9ef-e01e246f527a-viamrtsp-0_0_5-linux-arm64/bin/viamrtsp
/home/viam/.viam/packages/data/module/eacb3d64-4eab-43db-a9ef-e01e246f527a-viamrtsp-0_0_5-linux-arm64/bin/viamrtsp: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=30e4b5b9ecd4ba50ad2987e6429c0d85415eb144, for GNU/Linux 3.2.0, with debug_info, not stripped
```

.github/workflows/android-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        go-version: 1.21
+    
+    - name: Build viamrtsp binary and package module
+      run: |
+        TARGET_OS=android TARGET_ARCH=arm64 make module
+    
+    - name: Upload viamrtsp module to registry
+      uses: viamrobotics/upload-module@main
+      if: startsWith(github.ref, 'refs/tags/')
```
nicksanford, line 24: Why is this line needed?

seanavery: This checks for a tag event. (Had used it in previous modules).

The example wifi module uses `if: github.event_name == 'release'` (see[ here](https://github.com/viam-labs/wifi-sensor/blob/main/.github/workflows/deploy.yml#L29C7-L30C1)) -- but even this seems redundant with the on publish check.

seanavery: Ok, removed if check since we do not run it in publish.yml and should not be necessary.

nicksanford (approved):

PR #41, opened 2024-08-15
[RSDK-8197] [RSDK-8212]   Fix ViamRTSP Performance Regression using pool

hexbabe:
# Context

[RSDK-8197](https://viam.atlassian.net/browse/RSDK-8197)

Use custom pool allocation implementation to allocate a certain number of AVFrame blocks and reuse them when we can, rather than mallocing naively.

# Tests

### Registry 0.1.0 ver
Can never sustain 6 streams. CPU is fried (spikes to 95%+ all cores)
### This version
Can most of the time sustain 6 stable streams (it can more consistently do this in the mornings and after 4:30pm, so I assume Wifi). Streams will occasionally freeze but recover. No big CPU usage spikes, hovers between 60 and 80 percent all cores.

Video: https://viaminc.slack.com/archives/C06KBQEQDNC/p1725025942196499?thread_ts=1724778422.296799&cid=C06KBQEQDNC

Not too elucidating. But direct viewing stream comparisons show that the fix rivals the quality/stability of pre-regression.

## 8 streams on Jetson Orin Nano
Both 0.1.0 registry ver and this version can pretty consistently support and stream 8 rtsp components.

## Trying to make it segfault
Tried to make a segfault happen by switching resolutions mid-stream. Tried it 4 or 5 times. No segfault.

Also tried to switch settings drastically (e.g. change resolution, bitrate, I-frame interval, compression type) all in one go and nothing bad happened. Live stream changed and streamed with new settings.

## Profiler tests
https://viaminc.slack.com/archives/C06KBQEQDNC/p1724778422296799

## Debug logs
```
2024-08-30T14:59:13.668+0100	DEBUG	rdk.viamrtsp.rdk:component:camera/rtsp-1	host/pool.go:162	getCount: 349	{"log_ts": "2024-08-30T13:59:13.664Z"}
2024-08-30T14:59:13.672+0100	DEBUG	rdk.viamrtsp.rdk:component:camera/rtsp-1	host/pool.go:163	putCount: 345	{"log_ts": "2024-08-30T13:59:13.669Z"}
2024-08-30T14:59:13.677+0100	DEBUG	rdk.viamrtsp.rdk:component:camera/rtsp-1	host/pool.go:164	newCount: 32	{"log_ts": "2024-08-30T13:59:13.674Z"}
2024-08-30T14:59:13.682+0100	DEBUG	rdk.viamrtsp.rdk:component:camera/rtsp-1	host/pool.go:165	cleanCount: 0	{"log_ts": "2024-08-30T13:59:13.679Z"}
```

This was a live stream running for maybe ~10 seconds. `{x}Count` represents how many times pool method x was fully executed and the intended action was successfully performed (not just called) e.g. actually putting a frame back into the pool.


decoder.go
```diff
@@ -58,6 +61,26 @@ func (vc videoCodec) String() string {
 	}
 }
 
+type avFrameWrapper struct {
```
hexbabe, line 64: Need this because if we can't set a finalizer on C refs. Go's runtime would try to set a finalizer on a pointer that it doesn't recognize as a valid Go-allocated block and would fail

decoder.go
```diff
@@ -99,44 +122,54 @@ func newDecoder(codecID C.enum_AVCodecID, logger logging.Logger) (*decoder, erro
...
-func newH265Decoder(logger logging.Logger) (*decoder, error) {
-	return newDecoder(C.AV_CODEC_ID_H265, logger)
+func newH265Decoder(avFramePool *sync.Pool, logger logging.Logger) (*decoder, error) {
+	return newDecoder(C.AV_CODEC_ID_H265, avFramePool, logger)
 }
 
 // close closes the decoder.
 func (d *decoder) close() {
-	if d.dstFrame != nil {
-		C.av_frame_free(&d.dstFrame)
```
hexbabe, line 128: We should not manually free anymore now that we have the pool managing memory.

hexbabe: In fact, do we need to set to these to nil even? Is this extraneous?

decoder.go
```diff
@@ -146,60 +179,71 @@ func (d *decoder) decode(nalu []byte) (image.Image, error) {
 	avPacket.size = C.int(len(nalu))
 	res := C.avcodec_send_packet(d.codecCtx, &avPacket)
 	if res < 0 {
+		//nolint:nilnil // avcodec_send_packet raises non-critical errors
```
hexbabe, line 182: Something weird is happening with linter where it errors out on this nil nil now when it didn't before. Did we bump our lint deps or something?

hexbabe:
Hmm, I did some reading on finalizers and I think we need a better way to free our frames. Will do some more thinking today.

decoder.go
randhid: it always used to in my prs. I think it just ignored files with Cgo but may now include them. you can return an empty result struct if nil.

decoder.go
```diff
@@ -58,6 +61,26 @@ func (vc videoCodec) String() string {
...
+func allocateAVFrame() (*avFrameWrapper, error) {
+	avFrame := C.av_frame_alloc()
+	if avFrame == nil {
+		return nil, errors.New("failed to allocate AVFrame: out of memory or C libav internal error")
+	}
+	wrapper := &avFrameWrapper{frame: avFrame}
+	// Set a finalizer on the wrapper to ensure the C memory is freed
+	runtime.SetFinalizer(wrapper, func(w *avFrameWrapper) {
+		if w.frame != nil {
+			C.av_frame_free(&w.frame)
```
nicksanford, line 78: When would `w.frame` be nil?

rtsp.go
```diff
@@ -635,6 +641,19 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 		logger.Error(err.Error())
 		return nil, err
 	}
+
+	// initialize a pool to manage AVFrames so they don't bog down the garbage collector
+	framePool := &sync.Pool{
+		New: func() interface{} {
```
nicksanford, line 647:
```suggestion
		New: func() any{
```

rtsp.go
```diff
@@ -635,6 +641,19 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+			avFrame, err := allocateAVFrame()
+			if err != nil {
+				logger.Errorf("Failed to allocate AVFrame in frame pool: %v", err)
+				return nil
```
nicksanford, line 651: what happens if `nil` is returned from this function?

hexbabe:
Sorry for sending this out for review and then changing my mind on the implementation. I really don't like relying on runtime finalizer, so I think I will be changing the pool implementation to not use C pointers and memory at all and copy over bytes like we did in Sean P's fix, but combine that with pooling.

randhid:
> Sorry for sending this out for review and then changing my mind on the implementation. I really don't like relying on runtime finalizer, so I think I will be changing the pool implementation to not use C pointers and memory at all and copy over bytes like we did in Sean P's fix, but combine that with pooling.

That's okay. Do you want to close this so it's not noisy for Nick as you develop?

rtsp.go
```diff
@@ -798,12 +825,12 @@ func H2645StartCode() []byte {
...
 func (rc *rtspCamera) decodeAndStore(nalu []byte) error {
-	image, err := rc.rawDecoder.decode(nalu)
+	output, err := rc.rawDecoder.decode(nalu)
 	if err != nil {
 		return err
 	}
-	if image != nil {
-		rc.latestFrame.Store(&image)
+	if output != nil {
+		rc.latestOutput.Store(output)
```
nicksanford, line 833: I don't understand the interaction between the pool, and `latestOutput`.

rtsp.go
```diff
@@ -658,11 +678,18 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 	}
 	cancelCtx, cancel := context.WithCancel(context.Background())
 	reader := gostream.VideoReaderFunc(func(_ context.Context) (image.Image, func(), error) {
-		latest := rc.latestFrame.Load()
+		latest := rc.latestOutput.Load()
```
nicksanford, line 681: If we are calling `latestOutput.Store()` faster than we are calling `latestOutput.Load()` will we leak memory or do excessive garbage collection?

decoder.go
nicksanford, line 64: 1. How many `avFrameWrapper` do we expect to have for a given model instance?
2. When do we expect https://github.com/erh/viamrtsp/pull/41/files#diff-2a967f84f1af0d52111b8dbfbeadc44bfeb1092dacebb2bc0648a0e57284c04bR76 to run? AKA, when do we expect the `avFrameWrapper` to be garbage collected? How frequently does that happen?
3. Why is `allocateAVFrame` called both within pool.New and newDecoder? 
4. It seems like not all `allocateAVFrame` are managed by the pool, such as the one created in newDecoder. Why is that? How do I know if an avFrameWrapper is pool managed or not?

rtsp.go
```diff
@@ -495,7 +500,8 @@ func (rc *rtspCamera) initMJPEG(session *description.Session) error {
 			return
 		}
 
-		rc.latestFrame.Store(&img)
+		// manually set nil for avFramePtr since we don't use a pool for mjpeg frames
+		rc.latestOutput.Store(&imageAndPoolItem{img: img, poolItem: nil})
```
nicksanford, line 504: If we don't use a pool item for mjpeg frames... why is it in a type called `imageAndPoolItem`?
Can we please model this in a way that is less implicit & makes it clear which image are pool managed & which ones are not.

decoder.go
```diff
@@ -99,44 +122,54 @@ func newDecoder(codecID C.enum_AVCodecID, logger logging.Logger) (*decoder, erro
...
+	}
+
+	if d.codecCtx != nil {
+		C.avcodec_close(d.codecCtx)
+		d.codecCtx = nil
+	}
 }
 
-func (d *decoder) decode(nalu []byte) (image.Image, error) {
+func (d *decoder) decode(nalu []byte) (*imageAndPoolItem, error) {
```
nicksanford, line 172: why are we we returning a pointer here? 
this is the definition of imageAndPoolItem:
```
	imageAndPoolItem struct {
		img      image.Image
		poolItem *avFrameWrapper
	}
```

it has 2 pointers (image.Image is an interface so it's a pointer). Why are we returning a pointer to a struct that contains pointers? Lets just return the struct.

rtsp.go
```diff
@@ -658,11 +678,18 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
 	reader := gostream.VideoReaderFunc(func(_ context.Context) (image.Image, func(), error) {
-		latest := rc.latestFrame.Load()
+		latest := rc.latestOutput.Load()
 		if latest == nil {
 			return nil, func() {}, errors.New("no frame yet")
 		}
-		return *latest, func() {}, nil
+
+		poolCleanupCallback := func() {
+			if latest.poolItem != nil {
```
nicksanford, line 687: when is `poolItem` nil?

decoder.go
```diff
@@ -146,60 +179,71 @@ func (d *decoder) decode(nalu []byte) (image.Image, error) {
...
 			C.sws_freeContext(d.swsCtx)
 		}
 
-		d.dstFrame = C.av_frame_alloc()
-		d.dstFrame.format = C.AV_PIX_FMT_RGBA
-		d.dstFrame.width = d.srcFrame.width
-		d.dstFrame.height = d.srcFrame.height
-		d.dstFrame.color_range = C.AVCOL_RANGE_JPEG
-		res = C.av_frame_get_buffer(d.dstFrame, 1)
+		dstFrame, err := allocateAVFrame()
```
nicksanford, line 205: Why do we need to allocate here if we already got a frame from the pool here? https://github.com/erh/viamrtsp/pull/41/files#diff-2a967f84f1af0d52111b8dbfbeadc44bfeb1092dacebb2bc0648a0e57284c04bR194

decoder.go
```diff
@@ -146,60 +179,71 @@ func (d *decoder) decode(nalu []byte) (image.Image, error) {
...
 
+	dstFrameSize := C.av_image_get_buffer_size((int32)(d.dst.frame.format), d.dst.frame.width, d.dst.frame.height, 1)
+	dstFramePtr := (*[1 << 30]uint8)(unsafe.Pointer(d.dst.frame.data[0]))[:dstFrameSize:dstFrameSize]
+
 	// convert frame from YUV420 to RGB
-	res = C.sws_scale(d.swsCtx, frameData(d.srcFrame), frameLineSize(d.srcFrame),
-		0, d.srcFrame.height, frameData(d.dstFrame), frameLineSize(d.dstFrame))
+	res = C.sws_scale(d.swsCtx, frameData(d.src.frame), frameLineSize(d.src.frame),
+		0, d.src.frame.height, frameData(d.dst.frame), frameLineSize(d.dst.frame))
 	if res < 0 {
```
nicksanford, line 232: why did we change the order of `av_image_get_buffer_size` and `sws_scale`?

nicksanford (commented):
Sry, I already started reviewing before you closed the PR.

In general I agree that we should be able to reason about how big the pool is & when elements are removed and returned and when the pool grows.

I agree with your decision to move to another strategy that relies less / not at all on the golang GC.

dgottlieb (commented):
I don't have permissions to directly push a patch onto erh nor sean's repo. But [here's a patch](https://github.com/dgottlieb/viamrtsp/commit/df89935834b1352d00fbbd3418f53feeab7aeb9c) I've added documenting stuff that I had to figure out to understand the patch.

rtsp.go
hexbabe: I added Dan's documentation diff in a new commit, which should be elucidating

But basically the reason we went from `latestFrame image.Image` -> `latestOutput decoderOutput` is because we need the AVFrame in `rtsp.go`, specifically inside of`poolCleanupCallback` to `Put()` it back into the pool for reuse.

decoder.go
hexbabe: Shouldn't be. Removed

rtsp.go
hexbabe:
```
	d.dst = d.avFramePool.Get().(*avFrameWrapper)
	if d.dst == nil {
		return nil, errors.New("failed to obtain AVFrame from pool")
	}
```

We should catch it here. `.Get()` calls `.New()` when there are no existing, available items in the pool to return

rtsp.go
hexbabe: I changed the type name to `decoderOutput`

decoder.go
hexbabe: I added a lint ignore with a TODO attached to a ticket to fix it later

decoder.go
```diff
@@ -146,60 +179,71 @@ func (d *decoder) decode(nalu []byte) (image.Image, error) {
...
+		dstFrame, err := allocateAVFrame()
+		if err != nil {
+			return nil, errors.Errorf("AV frame allocation error while decoding: %v", err)
+		}
+		d.dst = dstFrame
+		d.dst.frame.format = C.AV_PIX_FMT_RGBA
+		d.dst.frame.width = d.src.frame.width
+		d.dst.frame.height = d.src.frame.height
+		d.dst.frame.color_range = C.AVCOL_RANGE_JPEG
+		res = C.av_frame_get_buffer(d.dst.frame, 1)
```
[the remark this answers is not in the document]
hexbabe: Yes that is my understanding

decoder.go
hexbabe: I think Dan's notes help a lot. Basically the frame we got from the pool may be a fresh blank frame with no metadata, or if the stream settings changed, a frame with the wrong metadata. That's why we reallocate here

decoder.go
```diff
@@ -21,11 +23,12 @@ import (
...
 // decoder is a generic FFmpeg decoder.
 type decoder struct {
-	logger   logging.Logger
-	codecCtx *C.AVCodecContext
-	srcFrame *C.AVFrame
-	swsCtx   *C.struct_SwsContext
-	dstFrame *C.AVFrame
+	logger      logging.Logger
+	codecCtx    *C.AVCodecContext
+	src         *avFrameWrapper
```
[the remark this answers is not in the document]
hexbabe: Will do

rtsp.go
hexbabe: No memory leak, but maybe GC

I think this would happen if something like:

Say I had a live stream going at 30fps, then I changed the frame rate in the Gost admin panel to 15fps. The equilibrium rate/number of items in the pool would be about halved then. So if half of those frames aren't being used anymore, they will be garbage collected properly during runtime by the finalizer. Whether that's excessive or not I am not entirely sure.

decoder.go
hexbabe: Updated. It is now just a struct returned

decoder.go
hexbabe: This was unintentional. Switching back

randhid:
Just double-checking we're looking at this approach for sure?

randhid:
> NGL I am not sure I am just responding to comments since they are good comments and questions
Alrightie. I'll re-review.

decoder.go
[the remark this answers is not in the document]
hexbabe: Done

decoder.go
```diff
@@ -58,6 +61,26 @@ func (vc videoCodec) String() string {
...
+
+// allocateAVFrame allocates a new AVFrame using C code with safety checks and returns the Go wrapper of it.
+func allocateAVFrame() (*avFrameWrapper, error) {
+	avFrame := C.av_frame_alloc()
+	if avFrame == nil {
+		return nil, errors.New("failed to allocate AVFrame: out of memory or C libav internal error")
+	}
+	wrapper := &avFrameWrapper{frame: avFrame}
+	// Set a finalizer on the wrapper to ensure the C memory is freed
+	runtime.SetFinalizer(wrapper, func(w *avFrameWrapper) {
```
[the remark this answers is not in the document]
hexbabe: Added a basic doc comment. Can expand it as I/we notice more cases

decoder.go
hexbabe: 1. Unfortunately, we can't directly check or log how many items are in a sync.Pool in Go. It's built to be a lightweight tool for handling temporary objects, so it doesn't keep track or expose how many items are in there.
2. As long as the stream settings don't change, we don't expect `avFrameWrapper`s to be garbage collected often i.e. the number of Get() is same/similar to the number of Put(). If there is a sudden decrease in the rate of Put() operations compared to Get(), it means that items are being retrieved from the pool faster than they are being returned, which can lead to a buildup of unused items in the pool. GC would then kick in. This could happen when we switch from a high frame rate to a lower one or a general slowdown in calling of `decode`/`onPacketRTP`.
3. I think 3 was answered [here](https://github.com/erh/viamrtsp/pull/41#discussion_r1720124104). 
4. I think all avFrameWrappers are eventually all pool managed. They are Put() into the pool through the `decode` -> `poolCleanupCallback` path. They are either freed via the finalizer garbage collection, or when stream settings changed and we manually free them with `d.dst.freeAVFrameWrapper()`.

hexbabe:
Haven't done any performance testing yet, but this is a draft of the custom pool implementation. It relies on the fixed logic introduced in https://github.com/viamrobotics/rdk/pull/4305

go.mod
```diff
@@ -9,14 +9,14 @@ require (
...
 	github.com/edaniels/golinters v0.0.5-0.20220906153528-641155550742
 	github.com/golangci/golangci-lint v1.57.2
-	github.com/pion/rtp v1.8.5
+	github.com/pion/rtp v1.8.7
 	github.com/pkg/errors v0.9.1
 	github.com/rhysd/actionlint v1.6.27
 	github.com/stretchr/testify v1.9.0
 	go.uber.org/zap v1.24.0
-	go.viam.com/rdk v0.26.0-rc0.0.20240503203304-30f601249ccf
+	go.viam.com/rdk v0.38.0
```
hexbabe, line 17: Bumping RDK to get https://github.com/viamrobotics/rdk/pull/4305

decoder.go
```diff
@@ -146,60 +179,71 @@ func (d *decoder) decode(nalu []byte) (image.Image, error) {
...
+	// get a frame from the pool
+	d.dst = d.avFramePool.Get().(*avFrameWrapper)
+	if d.dst == nil {
+		return nil, errors.New("failed to obtain AVFrame from pool")
+	}
 
+	// if frame size has changed, reinitialize dstFrame
+	if d.dst.frame.width != d.src.frame.width || d.dst.frame.height != d.src.frame.height {
 		if d.swsCtx != nil {
 			C.sws_freeContext(d.swsCtx)
```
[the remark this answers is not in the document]
hexbabe: https://viam.atlassian.net/browse/RSDK-8667?atlOrigin=eyJpIjoiMDVhYWNhODIyMzU5NGViMGFkYWE2NzRiOTc5NDM1ZGIiLCJwIjoiaiJ9

pool.go
```diff
@@ -0,0 +1,166 @@
...
+		select {
+		case <-ticker.C:
+			p.cleanup()
+		case <-p.stopCleaning:
+			return
+		}
+	}
+}
+
+func (p *framePool) cleanup() {
```
dgottlieb, line 122: Now that I've gotten to the bottom here, I see the purpose/justification for a lot of these other variables. Can we just delete this extra goroutine `cleanup` for now? And replace it with a simpler algorithm? Defined in the following comment.

I don't think the complexity is solving a real problem, but it is making the behavior/resource consumption non-deterministic.

For example: suppose we have `len(p.items)` = 11 and `p.preseededCount` is 10. Based on when the cleanupRoutine ticks, we could end up with (again, for example) 1 item leftover. Or 10.

And now if we never need to create a new pool item, the long-term size of `p.items` will never converge. So if there's no guarantee we'll ever shrink to only items with a `lastAccess` within the `p.maxAge`, why bother?

pool.go
```diff
@@ -0,0 +1,166 @@
...
+	p.logger.Debug("Trying to safely put frame back into pool.")
+
+	isFreed := frame.isFreed.Load()
+	isBeingServed := frame.isBeingServed.Load()
+	isInPool := frame.isInPool.Load()
+
+	if !isFreed && !isBeingServed && !isInPool {
+		p.put(frame)
+	} else {
+		p.logger.Debugf("Frame was already freed (%t) or is currently being served (%t) or is already in the pool (%t). Cannot put.",
```
dgottlieb, line 102: Use a `Warn` here

pool.go
```diff
@@ -0,0 +1,166 @@
...
+	p.items = p.items[1:]
+
+	p.logger.Debugf("Item was gotten from the pool. Len now: %d", len(p.items))
+	p.getCount++
+
+	item.frameWrapper.isInPool.Store(false)
+	return item.frameWrapper
+}
+
+func (p *framePool) put(frame *avFrameWrapper) {
```
dgottlieb, line 81: The surface area of this API is super small, this is fine -- not mandating a change here. 

But if users should* be calling `safelyPut` -- that should be named `put` to be more discoverable/the thing a user is most likely to type. And this can maybe be `doPut`? But honestly, I'd instead De Morgan the conditional in `safelyPut` and have the error case early-return. And I'd just move this logic inline with the "safely put" logic. Letting us delete this function.

pool.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		p.logger.Errorf("Failed to allocate AVFrame in frame pool: %v", err)
+		return nil
+	}
+	p.newCount++
+	return avFrame
+}
+
+func (p *framePool) get() *avFrameWrapper {
+	p.mu.RLock()
```
dgottlieb, line 64: The slicing on lines 71/72 is not safe to be done with 2 concurrent callers to `get`. This needs to be a "write lock". I expect the reader lock "works" here because `get` is only called from a single thread.

Given this is the only use of the "read"ing lock, just changing the `sync.RWMutex` to use a `sync.Mutex` is the right change to make.

pool.go
dgottlieb: The simpler algorithm being:
```
func [safely]Put(x item) {
  if len(p.items) >= p.preseededCount {
    x.frameWrapper.free()
  }

  p.items = append(p.items, x)
}
```

rtsp.go
```diff
@@ -438,16 +450,16 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 		}
 
 		for _, nalu := range au {
-			lastImage, err := rc.rawDecoder.decode(nalu)
+			output, err := rc.rawDecoder.decode(nalu)
```
dgottlieb, line 453: Dumb question -- definitely paged out a lot of stuff over the last couple weeks. In what circumstance does `err != nil` but `output == nil`? Is this related to that linting thing I said you ought not worry about?

rtsp.go
```diff
@@ -658,11 +675,26 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+		latest := rc.latestOutput.Load()
 		if latest == nil {
 			return nil, func() {}, errors.New("no frame yet")
 		}
-		return *latest, func() {}, nil
+
+		release := func() {}
+		// When model is mjpeg, latest frame wrapper will be nil.
+		if latest.frameWrapper != nil {
+			latest.frameWrapper.isBeingServed.Store(true)
```
dgottlieb, line 686: I'm pretty sure `isBeingServed` is equivalent to `isInPool`? This feels like an exercise where the user should remember to clal `isBeingServed(true)` after every `pool.get` and `isBeingServed(false)` prior to every `pool.put`. I don't think `isBeingServed`s access pattern will ever diverge from `isInPool`? But maybe I'm missing something.

dgottlieb (commented):
Leaving this feedback isolated to the pool for now. I haven't looked at whether the decoder logic changed since the last reviewed patch.

pool.go
```diff
@@ -0,0 +1,166 @@
...
+	mu           sync.RWMutex
+	stopCleaning chan struct{}
+	logger       logging.Logger
+}
+
+func newFramePool(maxAge, cleanupInterval time.Duration, preseededCount int, logger logging.Logger) *framePool {
+	pool := &framePool{
+		items:          make([]poolItem, 0, preseededCount),
+		maxAge:         maxAge,
+		preseededCount: preseededCount,
```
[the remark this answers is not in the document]
hexbabe: It is indeed like a target size that isn't super well enforced

pool.go
hexbabe: That's really nice I like that

pool.go
hexbabe: Action: split into different logs based on severity of violation

pool.go
hexbabe: Added

pool.go
```diff
@@ -0,0 +1,166 @@
...
+func (p *framePool) get() *avFrameWrapper {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	if len(p.items) == 0 {
+		return nil
+	}
+
+	item := p.items[0]
+	p.items = p.items[1:]
```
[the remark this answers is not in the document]
hexbabe: Added

pool.go
```diff
@@ -0,0 +1,166 @@
...
+	p.newCount++
+	return avFrame
+}
+
+func (p *framePool) get() *avFrameWrapper {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	if len(p.items) == 0 {
+		return nil
```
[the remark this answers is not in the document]
hexbabe: Combined `get` with `new` and deleted `new`

pool.go
hexbabe: Consolidated into `put`. Deleted `safelyPut`

pool.go
hexbabe: Done

rtsp.go
hexbabe: Added ref counting

hexbabe:
Made changes based on review and re-ran outlined manual tests for switching resolution/camera settings on RPI4 6 streams.

decoder.go
```diff
@@ -228,6 +240,7 @@ func (d *decoder) decode(nalu []byte) (*decoderOutput, error) {
 		// a fresh frame.
 		d.dst.free()
 		dstFrame, err := newAVFrameWrapper()
+		dstFrame.refs.Store(1)
```
dgottlieb, line 243: Use `Add(1)`. And if you're paranoid assert that "new" value is "1".

pool.go
```diff
@@ -2,165 +2,116 @@ package viamrtsp
...
-	if len(p.items) == 0 {
-		return nil
+	if len(p.frames) == 0 {
+		p.logger.Debug("No frames available in pool. Constructing new frame!")
+		frame, err := newAVFrameWrapper()
+		if err != nil {
+			p.logger.Errorf("Failed to allocate AVFrame in frame pool: %v", err)
+			return nil
+		}
+		frame.refs.Store(1)
```
dgottlieb, line 42: Same with using `Add(1)`.

pool.go
```diff
@@ -2,165 +2,116 @@ package viamrtsp
...
 
-func (p *framePool) cleanup() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
-	if len(p.items) <= p.preseededCount {
-		return
+	if len(p.frames) >= p.maxNumFrames {
+		// Cleanup logic: free the last item and truncate the slice
+		lastFrame := p.frames[len(p.frames)-1]
```
dgottlieb, line 87: What I'm reading here is:
* When a frame is deref'ed to a count of 0 and the pool is "full" is ready to use items
  * delete the last frame in the pool
  * and add the deref'ed frame back to the pool

Why do that dance of replacing the last pooled item with the "put" one? Instead of just free'ing the "put" one and calling it a day?

rtsp.go
```diff
@@ -675,26 +669,33 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+			// When the caller is done with the image, we return the AVFrame to the pool such
+			// that we can re-use its allocated byte buffer.
+			rc.logger.Debug("Release was called.")
+			frame.isBeingServed.Store(false)
+			frame.refs.Add(-1)
+			rc.avFramePool.put(frame)
 		}
 
-		return latest.img, release, nil
+		defer func() {
```
dgottlieb, line 694: Can we justify the defer as a comment?

rtsp.go
```diff
@@ -675,26 +669,33 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
-				rc.logger.Debug("Release was called.")
-				latest.frameWrapper.isBeingServed.Store(false)
-				rc.avFramePool.safelyPut(latest.frameWrapper)
-			}
+		release := func() {
+			// When the caller is done with the image, we return the AVFrame to the pool such
+			// that we can re-use its allocated byte buffer.
+			rc.logger.Debug("Release was called.")
+			frame.isBeingServed.Store(false)
+			frame.refs.Add(-1)
```
dgottlieb, line 690: `put` does the `Add(-1)` -- I think this is a double deref bug?

rtsp.go
```diff
@@ -675,26 +669,33 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+				return nil, func() {}, errors.New("no mjpeg frame yet")
+			}
+			return *img, func() {}, nil
 		}
 
-		release := func() {}
-		// When model is mjpeg, latest frame wrapper will be nil.
-		if latest.frameWrapper != nil {
-			latest.frameWrapper.isBeingServed.Store(true)
+		frame := rc.latestFrame.Load()
```
dgottlieb, line 680: I don't think the relationship between loading the latest frame and bumping the ref count is correct. What can happen when `handleLatestFrame` is called concurrently with this `VideoReaderFunc`?

rtsp.go
hexbabe: My understanding is that this `Add(-1)`'s equivalent `Add(1)` is inside of this `VideoReaderFunc` defer.

The one in `put`'s equivalent is in `get` (or when we make a new frame when the `srcYUV420Frame`/`dst` dimensions don't match).

rtsp.go
hexbabe: Sure does

`// Handle refs and flags after we pass the frame on to the server`

work?

pool.go
hexbabe: Good call; they are for sure equivalent

decoder.go
hexbabe: Got it

pool.go
hexbabe: Yep

rtsp.go
hexbabe: Thinking out loud here:

If `handleLatestFrame` grabs the underlying mutex first and swaps out the old frame for the new frame atomically, deref-ing and returning the old frame to the pool. This would cause `VideoReaderFunc` to wait for a sec then `load` to access the new frame— there should be no issues.

If the other way around happens, `VideoReaderFunc` loads in first, then `handleLatestFrame` swaps in a new value + derefs and puts the old one back in the pool, then said deref be part of a race condition with the ref in `VideoReaderFunc`.

If my understanding is correct, a possible solution would be to use a mutex for both critical sections? Are there alternatives I am not aware of

rtsp.go
dgottlieb: The defer is still executed before the return value is propagated to the server. So that wouldn't be accurate. Put another way, what's the consequence of not doing this in a defer?

hexbabe: Damn I misunderstood Golang. I thought deferred statements ran after when it really just guarantees The defer statement in Go essentially guarantees that certain logic will run at the very end of the function body, just before the function actually returns to its caller?

> what's the consequence of not doing this in a defer

Not much as long as we keep the lines right before the return. I guess the main difference and potential benefit would be that it is ran after all other deferred statements.

Would a better comment be 

`// Handle refs and flags right before handoff to server`

rtsp.go
hexbabe: Added a mutex

rtsp.go
hexbabe: My understanding was incorrect. Removed extra deref inside of `put`. In fact, there is now no ref handling at all in pool.go

rtsp.go
hexbabe: Action taken: removed those two lines from defer

pool.go
hexbabe: Done

pool.go
hexbabe: Done

decoder.go
hexbabe: Done

rtsp.go
```diff
@@ -658,11 +669,34 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+				return nil, func() {}, errors.New("no mjpeg frame yet")
+			}
+			return *img, func() {}, nil
+		}
+
+		if rc.latestFrame.Load() == nil {
 			return nil, func() {}, errors.New("no frame yet")
 		}
-		return *latest, func() {}, nil
+		frame := rc.latestFrame.Load()
```
dgottlieb, line 683: This still has the same bug -- a reader can get the latest frame, but before acquiring `frame.mu`, the "decoder" can deref the image -- freeing it.

This `latestFrame.Load` needs to be atomic with the incrementRefs. The (simplified) problem here is that there's a mutex on the frame object, but the important layer here is actually the `rc *rtspCamera`. Moving frames into/out of the `latestFrame` must be "atomic" with the reference counting for those frames.

hexbabe: Ok so IIUC, there are two concurrency problems

1. ref decrements need to be atomic with `put`-ing
2. Swapping out the `latestFrame` needs to be atomic with ref counting.

The per-frame mutex solves 1. We need something that does 2.

We can use an additional mutex maybe called `frameSwapMutex` owned by `rc` that makes 2 happen?

hexbabe: Went ahead and added it. Let me know if you think this is sufficient or the wrong direction etc

dgottlieb: > Frame ref decrements need to be atomic with put-ing

That's true for when updating the `latestFrame`, but not for calling the "release" function. I claim at least -- definitely would like you to convince yourself that claim is correct. Or come up with a counter-example for the claim.

> We can use an additional mutex maybe called frameSwapMutex owned by rc that makes 2 happen (additionally we could un-atomic the latestFrame pointer)?

That's right. We can also remove the per frame mutex in favor of atomics. What you have right now is correct in that the mutex protects concurrent access to the `refCount` variable. But I'm going to recommend using atomics here, 2 reasons:
* From a readability perspective, the current usage of the mutex makes it seem like it's needed for something.
* From a correctness perspective, because the lock is:
  * held in conjunction with the `frameSwapMutex` and
  * is held for longer than necessary
  * There's a risk of introducing a deadlock down the road.

rtsp.go
```diff
@@ -658,11 +672,39 @@ func newRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
...
+			// that we can re-use its allocated byte buffer.
+			frame.mu.Lock()
+			defer frame.mu.Unlock()
+
+			if refCount := frame.decrementRefs(); refCount == 0 {
+				rc.avFramePool.put(frame)
+			}
+		}
+
+		frame.incrementRefs()
```
dgottlieb, line 706: Because I left a comment directly on the comment, GH is forbidding me from submitting a review. So I'll say, let's move the `incrementRefs` up closer to the `frame := rc.latestFrame` assignment. That's when the `VideoReaderFunc` takes a stake in the lifetime of the frame. And it makes things more clear that the `frameSwapMu` is intended to be locked around that.

dgottlieb (changes requested):

rtsp.go
hexbabe: Got it. Done

rtsp.go
hexbabe: > That's true for when updating the latestFrame, but not for calling the "release" function. I claim at least -- definitely would like you to convince yourself that claim is correct. Or come up with a counter-example for the claim.

Because we aren't swapping in frame?

dgottlieb (approved):

hexbabe (commented):
Performance testing looks essentially the same as before Dan's review

PR #43, opened 2024-09-09
[RSDK-8589] erh -> viam

hexbabe:
https://viam.atlassian.net/browse/RSDK-8589

Something I want to make sure/clarify before merging and running `viam module update` with the new meta.json is how user configs will update from `erh:viamrtsp` to `viam:viamrtsp`

Do we need to run a migration?

Update: the answer is no, Michael Lee's logic includes old triplet -> new triplet redirection

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,201 @@
-MIT License
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Copyright (c) 2023 Viam
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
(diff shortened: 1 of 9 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 erh
+Copyright (c) 2023 Viam
```
hexbabe, line 3: Anything we need to do for the license? I noticed other camera modules have Apache 2.0

Why MIT?

randhid: He might have just added a random license, we should change this to Apache and change the date.

hexbabe: OK will swap out for Apache

hexbabe: Done

randhid (approved):

PR #44, opened 2024-09-09
[RSDK-7953] [RSDK-8709] Consolidate build workflow for all archs AND bump RDK

hexbabe:
[RSDK-7953](https://viam.atlassian.net/browse/RSDK-7953)

[RSDK-7953]: https://viam.atlassian.net/browse/RSDK-7953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/publish.yml
```diff
@@ -6,12 +6,44 @@ on:
...
+    runs-on: ${{ matrix.runner }}
+
     steps:
     - uses: actions/checkout@v3
-    - uses: viamrobotics/build-action@v1
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.21
```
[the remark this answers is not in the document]
hexbabe: Ok I can block on this ticket and figure that one out first

meta.json
```diff
@@ -21,10 +21,5 @@
       "model": "erh:viamrtsp:rtsp-mjpeg"
     }
   ],
-  "build": {
-    "build": "bash build.sh",
-    "path": "module.tar.gz",
-    "arch" : ["linux/amd64", "linux/arm64"]
```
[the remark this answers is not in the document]
hexbabe: No, the multiple arch targets here are being replaced with the matrix set up in the new `publish.yml`

.github/workflows/publish.yml
```diff
@@ -6,12 +6,44 @@ on:
 
 jobs:
   publish:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [linux/amd64, linux/arm64, android/arm64]
+        runner: [ubuntu-latest, buildjet-8vcpu-ubuntu-2204-arm]
```
hexbabe, line 12: Can technically remove these lists, but I think it's considered idiomatic to list out all the options in a list like this

build.sh
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
-set -e
-
-sudo apt-get update
-sudo apt-get install -y pkg-config
```
seanavery, line 5: Do we need to add `pkg-config` install in Makefile? 
Perhaps it already comes with ubuntu-latest.

seanavery (approved):
LGTM!

Do you think we need to test the publish flow in a private module first?

build.sh
hexbabe: I will test with private module and also try to confirm/deny this pkg-config hypothesis

hexbabe: https://github.com/hexbabe/sean-viamrtsp/actions/runs/10799689829/job/29956007233

The android and amd64 builds are working. The arm64 build does not work because my fork doesn't have buildjet. Testing now.

.github/workflows/publish.yml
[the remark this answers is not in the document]
hexbabe: Done

build.sh
hexbabe: I am unable to run the workflow in the viam-modules org because upstream secrets are not allowed to be pulled into PRs for security reasons

randhid (approved):

build.sh
seanavery: Sounds good

PR #48, opened 2024-09-13
[RSDK-7953]   Consolidate build workflow for all archs part 2

hexbabe:
If we were to continue using #44 only, we would lose support for Bullseye. Buildjet arm runners use a newer version of glibc. We should use canon's antique profile. This will also better align our manual test builds with deployed builds.

Update: I added a condition that doesn't use canon for Android. For some reason android builds don't work in the canon docker antique environment. The current android build flow does not use Docker or canon

Tests:
["Successful" linux/arm64 CI run](https://github.com/viam-modules/viamrtsp/actions/runs/10854614809/job/30125433385?pr=48)
["Successful" android/arm64 CI run](https://github.com/viam-modules/viamrtsp/actions/runs/10854576184/job/30125308560)
["Successful" linux/amd64 CI run](https://github.com/viam-modules/viamrtsp/actions/runs/10852087675/job/30117308055)

"successful" in quotes meaning that even though the CI technically failed, it failed on the upload step with a semver error, which is expected since we don't want them to actually upload yet.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-# on: [push, pull_request]
+on: [push, pull_request]
 
 jobs:
   build-and-test:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: [push, pull_request]
+# on: [push, pull_request]
 
 jobs:
   build-and-test:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: [push, pull_request]
+on: push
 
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -1,34 +1,54 @@
...
+    runs-on: ${{ matrix.runner }}
 
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.21.13
+        go-version: 1.23
```
hexbabe, line 41: canon requires 1.2.3

It's not the version we use to build the module (inside of canon)

.github/workflows/integration-test.yml
```diff
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: [push, pull_request]
+on: push
```
hexbabe, line 3: This causes the workflow to run twice. Just need `push` here

randhid (approved):

PR #49, opened 2024-09-17
[RSDK-8575]   Add better errors in decode.go

hexbabe:
[RSDK-8575](https://viam.atlassian.net/browse/RSDK-8575)

- [x] test in office

[RSDK-8575]: https://viam.atlassian.net/browse/RSDK-8575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: pull_request
+on: [push, pull_request]
 
 jobs:
   build-and-test:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: push
+on: pull_request
 
 jobs:
   build-and-test:
--- a/decoder.go
+++ b/decoder.go
@@ -60,21 +60,6 @@ func (vc videoCodec) String() string {
 	}
 }
 
-// recoverableError should be used when libav returns an error code, but we deem it non-severe,
-// and can choose to ignore/recover from it.
-type recoverableError struct {
-	Message string
-	Code    int
-}
-
-func (e *recoverableError) Error() string {
-	return fmt.Sprintf("libav error status: %s (code: %d)", e.Message, e.Code)
-}
-
-func newRecoverableAVError(code int, message string) *recoverableError {
-	return &recoverableError{Code: code, Message: message}
-}
-
 // avFrameWrapper wraps the libav AVFrame.
 type avFrameWrapper struct {
 	frame *C.AVFrame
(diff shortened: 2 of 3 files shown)


.github/workflows/integration-test.yml
```diff
@@ -1,6 +1,6 @@
 name: Integration test using mediamtx
 
-on: push
+on: [push, pull_request]
```
hexbabe, line 3: Sorry, changing this back. It triggered twice on branches made off of `viamrtsp` upstream, but not my forked branch. This makes it run on forked feature branches (we shouldn't be developing in the upstream repo anyways, my bad)

rtsp.go
```diff
@@ -457,6 +457,10 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 
 		for _, nalu := range au {
 			frame, err := rc.rawDecoder.decode(nalu)
+			var recoverableErr *RecoverableAVError
+			if errors.As(err, &recoverableErr) {
+				continue
```
randhid, line 462: Are there any differences between the go standard library's `errors.As` and the cursed one's? Cursed on is imported in this file.

randhid, line 462: Is there a continue forever case here? What's the worst thing that could happen if we never return from this loop?

randhid (changes requested):

decoder.go
```diff
@@ -208,15 +214,13 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
...
-		return nil, nil
+		return nil, newRecoverableAVError(int(res), "error sending packet to the decoder")
 	}
 
 	// receive frame if available
 	res = C.avcodec_receive_frame(d.codecCtx, d.src)
 	if res < 0 {
-		//nolint:nilnil // TODO RSDK-8575: change to not nil, nil
-		return nil, nil
+		return nil, newRecoverableAVError(int(res), "error receiving decoded frame from the decoder")
```
seanavery, line 223: Do we want to lookup FFmpeg error message (av_strerror) associated with res int?

decoder.go
```diff
@@ -60,6 +60,21 @@ func (vc videoCodec) String() string {
 	}
 }
 
+// RecoverableAVError should be used when libav returns an error code, but we deem it non-severe,
+// and can choose to ignore/recover from it.
+type RecoverableAVError struct {
+	Message string
+	Code    int
+}
```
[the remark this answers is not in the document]
hexbabe: Got it sry I'm _**still**_ getting used to Go and how capitalization is only used for exports

rtsp.go
hexbabe: Nah

Will remove the cursed pkg in https://viam.atlassian.net/browse/RSDK-8776

Can't do it as a quick one-liner since we have to refactor errors.wrap which is exclusive to the cursed pkg

decoder.go
```diff
@@ -60,6 +60,21 @@ func (vc videoCodec) String() string {
...
+
+func (e *RecoverableAVError) Error() string {
+	return fmt.Sprintf("libav error status: %s (code: %d)", e.Message, e.Code)
+}
+
+func newRecoverableAVError(code int, message string) *RecoverableAVError {
+	return &RecoverableAVError{Code: code, Message: message}
+}
```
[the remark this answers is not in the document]
hexbabe: Gotcha, now that it's a private struct do we still need constructor?

rtsp.go
```diff
@@ -457,6 +457,10 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
+			}
```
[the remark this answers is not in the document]
hexbabe: Done

decoder.go
hexbabe: Took a stab at implementing including av error strings everywhere they are provide-able

seanavery: Nice, LGTM

decoder.go
```diff
@@ -259,7 +291,7 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
 		// This allocates the underlying byte array to contain the image data.
 		res = C.av_frame_get_buffer(dst.frame, 1)
 		if res < 0 {
-			return nil, errors.New("av_frame_get_buffer() err")
+			return nil, newAvError(res, "av_frame_get_buffer() err")
```
seanavery, line 294: [q] Should we make this recoverable? Could there be some scenario where unable to allocate memory temporarily?

seanavery (approved):
LGTM. Thanks for adding a wrapper for fetching the FFmpeg error msg.

decoder.go
hexbabe: That makes sense to me. Could happen in memory pressure scenarios where we'd want to recover

hexbabe: Done

PR #50, opened 2024-09-18
[RSDK-8667]   Fix possible edge case for decode wonkiness using pool clear func

hexbabe:
https://viam.atlassian.net/browse/RSDK-8667

> This ticket was created to address Dan’s comment [here](https://github.com/erh/viamrtsp/pull/41#discussion_r1719998891)
> 
> Proposed solution:
> 
> Add an extra conditional where we detect mismatch between dst and source yuv frame that checks if the dst frame already has a defined non-zero height and width (a reused frame). If there is a non-zero dimension mismatch between dst and source yuv, then we know the resolution has changed.
> 
> We should flush the pool (possibly re-seed) to prevent the case Dan outlined with a fast resolution switch.

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -240,7 +240,8 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
 			dst.frame.width, dst.frame.height, d.src.width, d.src.height)
 
 		// Handle size changes while having previously initialized frames to avoid https://github.com/erh/viamrtsp/pull/41#discussion_r1719998891
-		if dst.frame.width > 0 || dst.frame.height > 0 {
+		frameWasPreviouslyInitialized := dst.frame.width > 0 && dst.frame.height > 0
+		if frameWasPreviouslyInitialized {
 			d.avFramePool.clear()
 			// Release old size frame
 			dst.free()
--- a/decoder.go
+++ b/decoder.go
@@ -64,7 +64,7 @@ func (vc videoCodec) String() string {
 type avFrameWrapper struct {
 	frame *C.AVFrame
 	// generation indicates which generation of frame formats the frame is on. It should only be set once and read from.
-	// It determines whether the pool can return to the pool or not.
+	// It determines whether the frame can return to the pool or not.
 	// If the generation matches that of the pool, then it can return, else it cannot. We need this because otherwise, when resolution changes
 	// occur, we would observe undefinied behavior. See https://github.com/erh/viamrtsp/pull/41#discussion_r1719998891
 	generation int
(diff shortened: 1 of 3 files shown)


decoder.go
```diff
@@ -238,19 +238,26 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
 	if dst.frame.width != d.src.width || dst.frame.height != d.src.height {
 		d.logger.Debugf("(re)making frame due to AVFrame dimension discrepancy: Dst (width: %d, height: %d) vs Src (width: %d, height: %d)",
 			dst.frame.width, dst.frame.height, d.src.width, d.src.height)
+
+		// Handle size changes while having previously initialized frames to avoid https://github.com/erh/viamrtsp/pull/41#discussion_r1719998891
+		frameWasPreviouslyInitialized := dst.frame.width > 0 && dst.frame.height > 0
+		if frameWasPreviouslyInitialized {
+			// Release previously initialized frames
+			d.avFramePool.clear()
```
dgottlieb, line 246: I think if `clear` freed all over the existing frames, this algorithm would be correct.

I don't think it can clear is guaranteed to erase all of the existing frames though. The `latestFrame` pointer can be released back into the pool after the clear is processed (technically, I do not believe this alone is sufficient for creating a bug with the edge case). Additionally, frames "borrowed" by `GetImage` will also eventually get released back into the pool (I do believe this is the case one can exploit to make the ABA bug manifest).

I expect we'll need to talk live to convince ourselves whether or not this patch as-is is complete in fixing the bug.

hexbabe: Dan and I talked in person

I pushed a draft of the `generation++` and changed `clear` to `clearAndStartNewGeneration`

decoder.go
```diff
@@ -63,6 +63,11 @@ func (vc videoCodec) String() string {
 // avFrameWrapper wraps the libav AVFrame.
 type avFrameWrapper struct {
 	frame *C.AVFrame
+	// generation indicates which generation of frame formats the frame is on. It should only be set once and read from.
+	// It determines whether the pool can return to the pool or not.
```
dgottlieb, line 67: It determines whether _the frame_ can return to the pool or not

decoder.go
```diff
@@ -114,7 +119,7 @@ func newAVFrameWrapper() (*avFrameWrapper, error) {
 	if avFrame == nil {
 		return nil, errors.New("failed to allocate AVFrame: out of memory or C libav internal error")
 	}
-	wrapper := &avFrameWrapper{frame: avFrame}
+	wrapper := &avFrameWrapper{frame: avFrame, generation: -1}
```
dgottlieb, line 122: You could consider having `newAVFrameWrapper` taking a generation as input. But I'm not going to block approval on this.

decoder.go
```diff
@@ -274,19 +279,26 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
...
 			dst.frame.width, dst.frame.height, d.src.width, d.src.height)
+
+		// Handle size changes while having previously initialized frames to avoid https://github.com/erh/viamrtsp/pull/41#discussion_r1719998891
+		frameWasPreviouslyInitialized := dst.frame.width > 0 && dst.frame.height > 0
+		if frameWasPreviouslyInitialized {
+			// Release previously initialized frames, and block old prev gen frames from returning to pool
+			dst.free()
+			d.avFramePool.clearAndStartNewGeneration()
+			// Make new frame to be initialized with new size
+			newDst, err := newAVFrameWrapper()
```
dgottlieb, line 290: IIUC, this `newDst.generation` is going to be `-1`.

I think that object will ultimately be returned to the pool in the future, but I don't see us ever updating its generation to the right value? Are we really not seeing [this panic](https://github.com/viam-modules/viamrtsp/pull/50/files#diff-5c0127b47636d901a18590597909d34f79c1e67840557675b5a981c26f7601e2R77)? I'm hoping I'm just missing something simple here.

hexbabe: you're right

just did some manual testing

sorry for sending this out for review before manual testing

dgottlieb (approved):
This all looks great to me!

PR #52, opened 2024-10-15
[RSDK-9000] Register Discovery callback and associated ONVIF stuff

hexbabe:
https://viam.atlassian.net/browse/RSDK-9000
- [x] Add more unit tests
- [x] Rebase `main` after https://github.com/viam-modules/viamrtsp/pull/53 is merged
- [x] Final sweep
- [x] Test with Flutter SDK (again)
- [x] Test with Golang SDK
- [x] Final final sweep


rtsp_test.go
```diff
@@ -254,64 +254,6 @@ func TestRTSPCameraPerformance(t *testing.T) {
 	})
 }
 
-func TestExtractXAddrsFromProbeMatch(t *testing.T) {
```
hexbabe, line 257: moved to internal package test file discovery_test.go

rtsp.go
```diff
@@ -61,10 +62,38 @@ var (
...
+					return nil, err
+				}
+				password, err := getStringFromExtra(extra, "password")
+				if err != nil {
+					return nil, err
+				}
+				camInfoList, err := viamonvif.DiscoverCameras(username, password, logger)
+				if err != nil {
+					return nil, err
+				}
```
randhid, line 94: should we log and still create if discover fails? Or will it necessarily mean that rtsp cameras are not on network and can't be reached?

randhid: Also, will this work for all models in this constructor?

randhid (changes requested):
Biggest thing is whether we want to call discover for all camera models in this module. Discovery logic + test looks good to me!

rtsp.go
hexbabe: > Also, will this work for all models in this constructor?
I think so since it's in the for each loop over all models registering

> should we log and still create if discover fails? Or will it necessarily mean that rtsp cameras are not on network and can't be reached?
Taking a look at `DiscoverCameras`, there are two returns
`return nil, fmt.Errorf("failed to retrieve network interfaces: %w", err)` and `return &CameraInfoList{Cameras: discoveredCameras}, nil`

So no error is blocking except for when we cannot discover the available network interfaces to call ws-discovery on, which feels like a big enough issue to just fail on I think

randhid: okay, it's an easy change if it is catastrophic for someone who knows their rtsp camera ids anyway, so we can change that behaviour with more feedback.

randhid: just to double check, the Discover function makes sense for the other models in this module, the `mjpeg`, the `h265` etc.

hexbabe: > just to double check, the Discover function makes sense for the other models in this module, the mjpeg, the h265 etc.

I think it does yeah. All the models we support require an rtsp address

seanavery (approved):
This is awesome! LGTM.

Just left a question about wether or not we need to check for RTP over UDP as well as TCP when fetching URIs.

viamonvif/discovery.go
```diff
@@ -0,0 +1,308 @@
...
+	var rtspUris []string
+	// Iterate over all profiles and get the RTSP stream URI for each one
+	for _, profile := range envelope.Body.GetProfilesResponse.Profiles {
+		logger.Debugf("Using profile token: %s", profile.Token)
+
+		getStreamURI := media.GetStreamUri{
+			StreamSetup: onvifxsd.StreamSetup{
+				Stream: onvifxsd.StreamType(streamTypeRTPUnicast),
+				Transport: onvifxsd.Transport{
+					Protocol: onvifxsd.TransportProtocol(transportProtocolRTSP),
```
seanavery, line 265: Just to clarify here:
https://github.com/use-go/onvif/blob/master/docs/wsdl/media2.wsdl#L679-L686

My understanding here is that we are checking for URIs that have `RTSP (RTP over TCP) support`.

Should we also check for `RtspUnicast (RTP over UDP) support`?

randhid (approved):

rtsp.go
hexbabe: > If Discovery errors, does RegisterComponent also fail?

No since its just being registered as a callback not being called

hexbabe:
@randhid @seanavery can you guys re-review https://github.com/viam-modules/viamrtsp/pull/52/commits/d12fbefdc88788278e8070da7d9144e4f1e7a7d8 ? I added some stuff after I realized that sometimes the wsdiscovery results are flaky i.e. about half the time, not all the available IPs are discovered on the first call. Solution was to make the call multiple times and compile the results

viamonvif/discovery.go
seanavery: Talked in IRL - does not look like we even need to add `Protocol` to query here.

RTSP over UDP vs TCP is another discussion all together. Does the viamrtsp `gortsplib` dep even support udp streaming? Not sure yet.

viamonvif/discovery.go
```diff
@@ -0,0 +1,354 @@
...
+				// other ws-discovery args
+				scopes := []string{}
+				types := []string{"dn:NetworkVideoTransmitter"}
+				namespaces := map[string]string{}
+				resp, err := wsdiscovery.SendProbe(iface.Name, scopes, types, namespaces)
+				if err != nil {
+					logger.Warnf("attempt %d: failed to send WS-Discovery probe on interface %s: %v", i+1, iface.Name, err)
+					break
+				}
+				discoveryResps = append(discoveryResps, resp...)
```
seanavery, line 130: [q] There are no issues with having duplicate responses here from multiple probes?

seanavery (approved):

randhid:
I looked at the commit and commented, but I don't know where they went, biggest things:
- you could use stoppable workers, your pattern matches what it does, but if a big refactor okay to do later.
- If we hit a length of 0 in any of your cases and never write to the results channel, do we deterministically hit the cancel and close functions? The only worry is that Discovery just hangs. It won't stop the constructor from being called, but could make the server wait or leak the routine on shutdown.

viamonvif/discovery.go
hexbabe: No duplicates. Checked manually and with AI

hexbabe:
> I looked at the commit and commented, but I don't know where they went, biggest things:
> 
> * you could use stoppable workers, your pattern matches what it does, but if a big refactor okay to do later.


Ah I tried using stoppable workers at first but they aren't as easily usable (perhaps this use case just isn't as documented) for non-"worker" type processes i.e. discovery since it doesn't run as a constant background process.

> * If we hit a length of 0 in any of your cases and never write to the results channel, do we deterministically hit the cancel and close functions? The only worry is that Discovery just hangs. It won't stop the constructor from being called, but could make the server wait or leak the routine on shutdown.
I'm at home right now and when we get nothing back on all interfaces, it does not hang. This is a really valid concern though and I'm gonna see if I can break it.

PR #56, opened 2024-11-08
[RSDK-9228] Default RTPPassthrough to true and update README

hexbabe:
https://viam.atlassian.net/browse/RSDK-9228

Manual test (h264)
Screenshot 2024-11-11 at 12 46 46 PM
As you can see, it's not specified in the config, but in the logs it indicates rtp passthrough is happening

Also verified that setting the field true results in passthrough still happening, and setting the field to false results in passthrough not happening

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The following attributes are available for all models of `viamrtsp` cameras:
 | Name    | Type   | Inclusion    | Description |
 | ------- | ------ | ------------ | ----------- |
 | `rtsp_address` | string | **Required** | The RTSP address where the camera streams. |
-| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec if this attribute is set to `true`. <br> Default: `false` |
+| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec. It will be on by default. Set to false to bypass H264 RTP passthrough. <br> Default: `true` |
 
 ### Example configuration
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The following attributes are available for all models of `viamrtsp` cameras:
 | Name    | Type   | Inclusion    | Description |
 | ------- | ------ | ------------ | ----------- |
 | `rtsp_address` | string | **Required** | The RTSP address where the camera streams. |
-| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec. It will be on by default. Set to false to bypass H264 RTP passthrough. <br> Default: `true` |
+| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec. It will be on by default. Set to false to disable H264 RTP passthrough. <br> Default: `true` |
 
 ### Example configuration
 
--- a/rtsp.go
+++ b/rtsp.go
@@ -101,7 +101,7 @@ func init() {
 // Config are the config attributes for an RTSP camera model.
 type Config struct {
 	Address          string                             `json:"rtsp_address"`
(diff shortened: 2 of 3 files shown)


hexbabe:
@nicksanford adding you since you suggested the change to use *bool and I want to make sure that I'm doing it right from your pov

rtsp.go
```diff
@@ -693,11 +693,16 @@ func NewRTSPCamera(ctx context.Context, _ resource.Dependencies, conf resource.C
 
 	framePool := newFramePool(initialFramePoolSize, logger)
 
+	rtpPassthrough := true
+	if newConf.RTPPassthrough != nil {
+		rtpPassthrough = *newConf.RTPPassthrough
+	}
+
```
randhid, line 700: Isn't this constructor shared by the other models that do not support rtp passthrough as well? mjpeg, h265?

randhid (changes requested):

rtsp.go
hexbabe: Yeah it is I think it's handled on a lower level when it bypasses passthrough after it checks that the properties of the stream isn't h264

README.md
```diff
@@ -30,7 +30,7 @@ The following attributes are available for all models of `viamrtsp` cameras:
 | Name    | Type   | Inclusion    | Description |
 | ------- | ------ | ------------ | ----------- |
 | `rtsp_address` | string | **Required** | The RTSP address where the camera streams. |
-| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec if this attribute is set to `true`. <br> Default: `false` |
+| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec. It will be on by default. Set to false to bypass H264 RTP passthrough. <br> Default: `true` |
```
nicksanford, line 33:
```suggestion
| `rtp_passthrough` | bool | Optional | RTP passthrough mode (which improves video streaming efficiency) is supported with the H264 codec. It will be on by default. Set to false to disable H264 RTP passthrough. <br> Default: `true` |
```

rtsp_test.go
```diff
@@ -94,7 +94,8 @@ func TestRTSPCamera(t *testing.T) {
 				timeoutCtx, timeoutCancel := context.WithTimeout(context.Background(), time.Second*10)
 				defer timeoutCancel()
 				config := resource.NewEmptyConfig(camera.Named("foo"), ModelAgnostic)
-				config.ConvertedAttributes = &Config{Address: "rtsp://" + h.s.RTSPAddress, RTPPassthrough: true}
+				shouldPassthrough := true
+				config.ConvertedAttributes = &Config{Address: "rtsp://" + h.s.RTSPAddress, RTPPassthrough: &shouldPassthrough}
```
nicksanford, line 98: we need a test that shows that `&Config{Address: "rtsp://" + h.s.RTSPAddress}` results in `when RTPPassthrough = true` behavior.

nicksanford:
have you tested rtp_passthrough (not set, on, off) on all models?

rtsp_test.go
hexbabe: Added nil case, true case, false case in a loop

hexbabe:
> have you tested rtp_passthrough (not set, on, off) on all models?

`viamrtsp sean-viamrtsp/rtsp.go:518 rtp_passthrough is only supported for H264 codec. rtp_passthrough features disabled due to MJPEG RTSP track log_ts 2024-11-12T17:12:57.817Z`

`viamrtsp host/rtsp.go:430 rtp_passthrough is only supported for H264 codec. rtp_passthrough features disabled due to H265 RTSP track log_ts 2024-11-`

Got these logs when nil and false

Always got logs for `rdk.modmanager camera/client.go:425 Error subscribing to RTP. Closing passthrough buffer.` for both h265 and mjpeg codecs

randhid (approved):

PR #57, opened 2024-11-08
Don't use canon in CI

hexbabe:
[0.2.0-rc1](https://github.com/viam-modules/viamrtsp/releases/tag/0.2.0-rc1) does not work on bullseye. This is likely because `canon` is being annoying in the CI runner and the artifact was not actually built inside `viam-rtsp-antique`. The workaround is to directly use docker from the source RDK build images

Added new debugging step to make sure the module.tar.gz successfully builds

Consolidated build module logic in CI to use a single DRY reusable action `build-module.yml` that builds it in the actual build container used to publish modules in to unite build and test environments


randhid:
Subbing martha for myself so we get more build action knowledge across module-heavy teams.

.github/workflows/integration-test.yml
```diff
@@ -29,11 +37,13 @@ jobs:
...
-    - name: Set up Go
-      uses: actions/setup-go@v2
+    - name: Build module
+      uses: ./.github/actions/build-module
       with:
-        go-version: 1.21.13
-    
+        target_os: linux
+        target_arch: ${{ matrix.platform.arch }}
+        docker_image: ${{ matrix.platform.docker_image }}
```
martha-johnston, line 45: you don't still need the go version? does docker already use a specific version so you don't need it here?

.github/workflows/publish.yml
```diff
@@ -32,25 +33,34 @@ jobs:
...
+          echo "module.tar.gz exists"
+        else
+          echo "module.tar.gz does not exist"
+          exit 1
         fi
-        TARGET_OS=${{ matrix.target_os }} TARGET_ARCH=${{ matrix.target_arch }} make module
 
     - name: Upload viamrtsp module to registry
-      uses: viamrobotics/upload-module@main
+      uses: viamrobotics/upload-module@v1
```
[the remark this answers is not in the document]
martha-johnston: good to know, I've been using main

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,44 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.21.13
+
+    - name: Pull Docker image
+      run: docker pull ${{ inputs.docker_image }}
```
martha-johnston, line 22: so using with: and including a docker image in the other workflows makes that an input here? i'm not familiar with inputs

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,44 @@
...
+          sh -c "make module"
+      shell: bash
+
+    - name: Verify module.tar.gz exists recursively
+      run: |
+        if find . -name module.tar.gz | grep -q .; then
+          echo "module.tar.gz exists"
+        else
+          echo "module.tar.gz does not exist"
+          exit 1
```
martha-johnston, line 42: so the goal here is just to report if module.tar.gz exists? will this cause the workflow to fail?

.github/workflows/android-build-test.yml
```diff
@@ -14,11 +14,9 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Set up Go
-      uses: actions/setup-go@v2
```
martha-johnston, line 18: will there be issues that you were using `setup-go@v2` and now in `action.yml` you're using `setup-go@v3`?

.github/actions/build-module/action.yml
hexbabe: yeah I think of it as params and args

params being what's defined above in `inputs:`

and `with:` being the args passed along

.github/actions/build-module/action.yml
hexbabe: I originally had this in for debugging but decided to keep it since it made sanity checking the existence of the `make module` `module.tar.gz` artifact easier. I think we can remove though it's not critical

.github/workflows/android-build-test.yml
hexbabe: good point we don't need to bump it

.github/workflows/integration-test.yml
hexbabe: I think because we're `  using: 'composite'` in the action.yml, the setup go action carries over to the environment of the main workflow

martha-johnston:
I’m still not _super_ knowledgeable about these workflows, so I’d say def check on it after you merge to make sure there’s no issues, but as far as I can tell, looks good

seanavery (approved):
LGTM!

Nothing to add here - really like the consolidated build flow :)

PR #58, opened 2024-11-11
upnp to find ip of a camera

erh:


viamupnp/find.go
```diff
@@ -0,0 +1,165 @@
...
+	} `xml:"specVersion"`
+	Device struct {
+		Manufacturer string `xml:"manufacturer"`
+		ModelName    string `xml:"modelName"`
+		SerialNumber string `xml:"serialNumber"`
+	} `xml:"device"`
+}
+
+func readDeviceDesc(ctx context.Context, url string) (*deviceDesc, error) {
+	cli := &http.Client{}
```
hexbabe, line 133: [nit] this will create a new http client per call. Should be okay though since this code path is run at most once per (re)configure?

hexbabe:
```
// near top of file
const httpTimeout = 10 * time.Second

// blah blah blah

client := &http.Client{
    Timeout: httpTimeout,
}
```

Might also be best practice to add a timeout

hexbabe (changes requested):
Broad question: Is this what we're going with instead of a wrapper module? Or is that still a separate thing

rtsp.go
```diff
@@ -132,6 +135,33 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
+		return nil, err
+	}
+
+	pcs := strings.Split(u.Host, ":")
+	if len(pcs) > 2 { //nolint:mnd
+		return nil, fmt.Errorf("invalid url [%s]", conf.Address)
+	}
+
+	if pcs[0] == "UPNP_DISCOVER" {
+		host, err := viamupnp.FindHost(ctx, logger, conf.Query)
```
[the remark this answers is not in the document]
erh: did something.

erh:
> Broad question: Is this what we're going with instead of a wrapper module? Or is that still a separate thing

getting rid of wrapper idea for now

PR #69, opened 2025-01-13
[RSDK-9688, RSDK-9732, RSDK-9733] Use libav for JPEG encoding

seanavery:
## Description

This PR moves to using libAV for JPEG mime handling and streamlines the frame pipeline from decoding to API returns from Image.
- Removes RGB pixel conversion in decoder.
  - We now simply copy yuv420 bytes to the AvFrameWrapper in the pool. 
  - Moves AvFrameWrapper toImage helper to fill Ycbcr image.
- Replaces golang jpeg encoder with libav MJPEG codec.
  - Sets up an MJPEG encoder that will reinitiate on size changes.
- Makes the MJPEG decoder path more efficient by removing going to and from a go image.

## Tests
- basic tests:
  - h264 Image ✅ 
  - h264 passthrough ✅ 
  - h265 Image ✅ 
  - mjpeg Image ✅ 
- frame size changes while running ✅ 
- two concurrent clients polling Image ✅ 
- reconfigure stuff:
  - module reconfigure ✅ 
  - viam-server shutdown does not hang ✅ 
  - removing and adding module no hangs ✅ 

## Performance
- CPU Profiling
  - Noticing slightly less time spent in performance iteration (10-15%).
    - This is to be expected since we replaced a pixel conversion with a memory copy into the pool and removed going to a go image before encoding.
[cpu-profile-main-branch.txt](https://github.com/user-attachments/files/18414369/cpu-profile-main-branch.txt)
[cpu-profile-feature-branch.txt](https://github.com/user-attachments/files/18414393/cpu-profile-feature-branch.txt)

- Memory Profiling 
  - As expected, no appreciable diff in memory
[mem-profile-main-branch.txt](https://github.com/user-attachments/files/18414930/mem-profile-main-branch.txt)
[mem-profile-feature-branch.txt](https://github.com/user-attachments/files/18414450/mem-profile-feature-branch.txt)

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -349,6 +349,7 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
 	// Copy the frame properties from the source frame to the destination frame.
 	// This will fill fields not explicitly set in the initial dst frame allocation.
 	if res := C.av_frame_copy_props(dst.frame, d.src); res < 0 {
+		// We should never reach this point if av_frame_copy() succeeded.
 		return nil, newAvError(res, "av_frame_copy_props() failed")
 	}
 
--- a/decoder.go
+++ b/decoder.go
@@ -19,6 +19,10 @@ import (
 	"go.viam.com/rdk/logging"
 )
 
+const (
+	yuv420SubsampleRatio = 2
+)
+
 // decoder is a generic FFmpeg decoder.
 type decoder struct {
 	logger   logging.Logger
@@ -102,8 +106,7 @@ func (w *avFrameWrapper) free() {
 
 // toImageYCbCr maps the underlying AVFrame (in YUV420P format) to a Go image.YCbCr.
 // Returns nil if frame is not AV_PIX_FMT_YUV420P or if something goes wrong.
-func (w *avFrameWrapper) toImageYCbCr() image.Image {
-	// Ensure the frame format is YUV420P.
+func (w *avFrameWrapper) toImage() image.Image {
 	if w.frame.format != C.AV_PIX_FMT_YUV420P {
 		return nil
 	}
@@ -111,7 +114,6 @@ func (w *avFrameWrapper) toImageYCbCr() image.Image {
 	width := int(w.frame.width)
 	height := int(w.frame.height)
 
-	// Strides (a.k.a. linesize in FFmpeg).
 	// For YUV420P:
 	//   - data[0] = Y plane, linesize[0] = stride for Y
 	//   - data[1] = U plane, linesize[1] = stride for U
@@ -123,21 +125,17 @@ func (w *avFrameWrapper) toImageYCbCr() image.Image {
 	// Y plane is full resolution: height * yStride
 	// U and V planes are half resolution in both dimensions: (height/2) * cStride
 	yPlaneSize := yStride * height
-	cPlaneSize := cStride * (height / 2)
+	cPlaneSize := cStride * (height / yuv420SubsampleRatio)
 
-	// Map the Y plane into a Go slice.
 	yDataPtr := unsafe.Pointer(w.frame.data[0])
 	ySlice := (*[1 << 30]byte)(yDataPtr)[:yPlaneSize:yPlaneSize]
 
-	// Map the U (Cb) plane into a Go slice.
 	cbDataPtr := unsafe.Pointer(w.frame.data[1])
 	cbSlice := (*[1 << 30]byte)(cbDataPtr)[:cPlaneSize:cPlaneSize]
 
-	// Map the V (Cr) plane into a Go slice.
 	crDataPtr := unsafe.Pointer(w.frame.data[2])
 	crSlice := (*[1 << 30]byte)(crDataPtr)[:cPlaneSize:cPlaneSize]
 
-	// Construct the Go image.YCbCr. For YUV420p, the subsample ratio is 4:2:0.
 	ycbcr := &image.YCbCr{
 		Y:              ySlice,
 		Cb:             cbSlice,
--- a/decoder.go
+++ b/decoder.go
@@ -312,9 +312,10 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
 
 	// If the frame from the pool has the wrong size, (re-)initialize it.
 	if dst.frame.width != d.src.width || dst.frame.height != d.src.height || dst.frame.format != d.src.format {
-		d.logger.Debugf("(re)making frame due to AVFrame discrepancy: Dst (width: %d, height: %d, format: %d) vs Src (width: %d, height: %d, format: %d)",
-			dst.frame.width, dst.frame.height, dst.frame.format, d.src.width, d.src.height, d.src.format)
-
+		d.logger.Debugf("(re)making frame due to AVFrame discrepancy: "+
+			"Dst (width: %d, height: %d, format: %d) vs Src (width: %d, height: %d, format: %d)",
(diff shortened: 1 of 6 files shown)


rtsp.go
```diff
@@ -2,12 +2,9 @@
 package viamrtsp
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
-	"image"
-	"image/jpeg"
```
randhid, line 10: 🥳

rtsp.go
```diff
@@ -913,41 +910,39 @@ func isCompactableH264(nalu []byte) bool {
...
-		return nil, camera.ImageMetadata{}, errors.New("no mjpeg frame yet")
+	if videoCodec(rc.currentCodec.Load()) == MJPEG {
+		mjpegBytes := rc.latestMJPEGBytes.Load()
+		if mjpegBytes == nil {
+			return nil, camera.ImageMetadata{}, errors.New("no frame yet")
+		}
+		return *mjpegBytes, camera.ImageMetadata{
+			MimeType: rutils.MimeTypeJPEG,
+		}, nil
 	}
```
randhid, line 926: This is where we'd add Yuv frame handling?

randhid: Let's make it a case structure from the onset if so

mime.go
```diff
@@ -1,20 +1,114 @@
...
+	return dataGo, camera.ImageMetadata{
 		MimeType: rutils.MimeTypeJPEG,
 	}, nil
 }
+
+func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
+	// Lock to prevent modifying encoder while it is being used concurrently.
+	// Frame param changes are rare, so we can afford to block here.
+	mh.mu.Lock()
+	defer mh.mu.Unlock()
```
randhid, line 75: Is there a way to test this with reconfiguring and rebuilding the camera? Either in our test suite in this module, or verifying that there is no hang or orphaned routines when we reconfigure the model, restarting the module or viam-server.

mime.go
```diff
@@ -1,20 +1,114 @@
...
+	codec := C.avcodec_find_encoder(C.AV_CODEC_ID_MJPEG)
+	if codec == nil {
+		return errors.New("failed to find MJPEG encoder")
+	}
+	mh.jpegEnc = C.avcodec_alloc_context3(codec)
+	mh.jpegEnc.width = frame.width
+	mh.jpegEnc.height = frame.height
+	mh.jpegEnc.pix_fmt = C.AV_PIX_FMT_YUVJ420P
+	// We don't care about accurate timestamps still frames
+	mh.jpegEnc.time_base = C.AVRational{num: 1, den: 1}
```
randhid, line 89: Do we have appropriate nil checks in place so far for everything we may access from the jpeg encoder? How easy is it to write tests for this file as it stands right now?

randhid (changes requested):
Done with first pass - I have a few questions I'd like answered and some where I'd like verification that zeros/nils are handled appropriately. Also is there any way to ad more testing to this pr? Or is it too cgo heavy and appropriate for a secondary ticket?

decoder.go
```diff
@@ -102,18 +104,49 @@ func (w *avFrameWrapper) free() {
 	}
 }
 
-// toImage takes the underlying av frame and embeds it into a RGBA image struct.
+// toImageYCbCr maps the underlying AVFrame (in YUV420P format) to a Go image.YCbCr.
```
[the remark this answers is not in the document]
seanavery: thanks!

decoder.go
```diff
@@ -102,18 +104,49 @@ func (w *avFrameWrapper) free() {
...
-		Stride: bytesPerPixel * (int)(w.frame.width),
-		Rect: image.Rectangle{
-			Max: image.Point{(int)(w.frame.width), (int)(w.frame.height)},
-		},
+	if w.frame.format != C.AV_PIX_FMT_YUV420P && w.frame.format != C.AV_PIX_FMT_YUVJ420P {
+		return nil
 	}
+
+	width := int(w.frame.width)
+	height := int(w.frame.height)
```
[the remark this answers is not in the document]
seanavery: I dont think so but added check to be safe.

decoder.go
```diff
@@ -102,18 +104,49 @@ func (w *avFrameWrapper) free() {
...
+		Y:              ySlice,
+		Cb:             cbSlice,
+		Cr:             crSlice,
+		YStride:        yStride,
+		CStride:        cStride,
+		SubsampleRatio: image.YCbCrSubsampleRatio420,
+		Rect:           image.Rect(0, 0, width, height),
+	}
+
+	return ycbcr
```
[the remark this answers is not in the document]
seanavery: sounds good - changed to directly return struct.

decoder.go
```diff
@@ -296,37 +329,27 @@ func (d *decoder) decode(nalu []byte) (*avFrameWrapper, error) {
...
 
-	// convert frame from YUV420 to RGB
-	res = C.sws_scale(d.swsCtx, frameData(d.src), frameLineSize(d.src),
-		0, d.src.height, frameData(dst.frame), frameLineSize(dst.frame))
-	if res < 0 {
-		return nil, newAvError(res, "sws_scale() err")
+	// Copy the frame properties from the source frame to the destination frame.
+	// This will fill fields not explicitly set in the initial dst frame allocation.
+	if res := C.av_frame_copy_props(dst.frame, d.src); res < 0 {
+		return nil, newAvError(res, "av_frame_copy_props() failed")
```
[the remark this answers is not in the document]
seanavery: `av_frame_copy` just copies over the pixel data, the `av_frame_copy_props` will copy over all of the other metadata see struct [here](https://ffmpeg.org/doxygen/7.0/structAVFrame.html). We want to make sure that everything is filled for the downstream mime conversions using libav (mjpeg encode and other pixel conversions we will support).

If the frame_copy does not error l I do not see a realistic case where the copy_props then errors.. But can do some more testing to see if we really need to fail on copy_props failure.

rtsp.go
seanavery: ok added switch case template.

mime.go
seanavery: Manually tested reconfigure and we are looking good. We have a `resource.AlwaysRebuild` flag so should be pretty safe here.

decoder.go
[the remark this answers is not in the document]
randhid: That's okay - add a comment above this line that we're not expecting this to ever be hit if the `av_frame_copy` fails.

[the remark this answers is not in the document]
seanavery: Sounds good, added comment.

mime.go
seanavery: Added unit tests for jpeg conversion mime handling in `mime_test.go` & `mime_cgo.go`.

mime_test.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+
+	t.Run("invalid frame fails", func(t *testing.T) {
+		frame := createInvalidFrame()
+		test.That(t, frame, test.ShouldNotBeNil)
+		defer freeFrame(frame)
+		logger := logging.NewDebugLogger("mime_test")
+		mh := newMimeHandler(logger)
+		bytes, metadata, err := mh.convertJPEG(frame)
+		test.That(t, err, test.ShouldNotBeNil)
```
randhid, line 48: Check that it is the error you're expecting.

seanavery: Good call added string contains check.

randhid (approved):

PR #71, opened 2025-01-17
[RSDK-9700] - Add YUYV mimetype handling

seanavery:


mime.go
```diff
@@ -107,8 +120,96 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+	if mh.yuyvSwsCtx == nil {
+		return nil, camera.ImageMetadata{}, errors.New("failed to create YUYV converter")
+	}
+	if mh.yuyvFrame == nil {
+		return nil, camera.ImageMetadata{}, errors.New("failed to create yuyv destination frame")
+	}
+	res := C.sws_scale(
+		mh.yuyvSwsCtx,
+		(**C.uint8_t)(unsafe.Pointer(&frame.data[0])),
+		(*C.int)(unsafe.Pointer(&frame.linesize[0])),
```
randhid, line 138: does a nil check on frame exist elsewhere?

mime.go
```diff
@@ -107,8 +120,96 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
 	return nil
 }
 
+func (mh *mimeHandler) convertYUYV(frame *C.AVFrame) ([]byte, camera.ImageMetadata, error) {
```
randhid, line 123: I notice no locks in this function; if it is meant only to be called from other functions handling a mutex lock, add a comment above it saying so.

mime.go
```diff
@@ -107,8 +120,96 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+
+	return yuyvPacket, camera.ImageMetadata{
+		MimeType: mimeTypeYUYV,
+	}, nil
+}
+
+func (mh *mimeHandler) initYUYVContext(frame *C.AVFrame) error {
+	mh.mu.Lock()
+	defer mh.mu.Unlock()
+	mh.logger.Info("creating YUYV sws cosntext with frame size: ", frame.width, "x", frame.height)
```
randhid, line 163:
```suggestion
	mh.logger.Info("creating YUYV sws context with frame size: ", frame.width, "x", frame.height)
```

mime.go
```diff
@@ -107,8 +120,96 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+	if mh.yuyvFrame != nil {
+		C.av_frame_free(&mh.yuyvFrame)
+	}
+	mh.yuyvFrame = C.av_frame_alloc()
+	if mh.yuyvFrame == nil {
+		return errors.New("failed to allocate YUYV frame")
+	}
+	mh.yuyvFrame.width = frame.width
+	mh.yuyvFrame.height = frame.height
+	mh.yuyvFrame.format = C.AV_PIX_FMT_YUYV422
```
randhid, line 176: do we ever need to care about yuyv420 differences here?

randhid (approved):
LGTM mod some nits. Good job Sean

mime.go
seanavery: Good call. Added nil checks to convert methods.

mime.go
seanavery: Ty. Fixed sp.

mime.go
seanavery: The output of the sws_scale operation here is always `AV_PIX_FMT_YUYV422`. We can handle multiple input formats (`YUV420P` or `YUVJ420P`) which is automatically handled in the decoder to avframe pool pipeline.

mime.go
randhid: Please be sure to test after for hangs on module restart/closure.

PR #73, opened 2025-01-22
add onvif discovery service

JohnN193:
https://viam.atlassian.net/browse/RSDK-9625

adds the onvif discover service. Made it a separate package from the current `viamonvif` to avoid circular dependencies happening and to not break the current discovery workflow that people use. 

requires this [rdk pr](https://github.com/viamrobotics/rdk/pull/4697) to be released. In addition only viam-servers on this version or later can use the module due to the tagger/proto reflection issue we ran into.

[test robot/config](https://app.viam.com/machine/dc201dda-4707-423b-8b6d-914c471a2743/configure/builder)

example code(get the API key from the [robot](https://app.viam.com/machine/dc201dda-4707-423b-8b6d-914c471a2743/connect/code-sample/golang))
```
package main

import (
	"context"

	"go.viam.com/rdk/logging"
	"go.viam.com/rdk/robot/client"
	"go.viam.com/rdk/services/discovery"
	"go.viam.com/utils/rpc"
)

func main() {
	logger := logging.NewLogger("client")
	machine, err := client.New(
		context.Background(),
		"rtsp-cams-main.ytxdza0q92.viam.cloud",
		logger,
		client.WithDialOptions(rpc.WithEntityCredentials( 
            /* Replace "<API-KEY-ID>" (including brackets) with your machine's api key id */
			"<API-KEY-ID>",
			rpc.Credentials{
				Type:    rpc.CredentialsTypeAPIKey, 
                /* Replace "<API-KEY>" (including brackets) with your machine's api key */
				Payload: "<API-KEY>",
			})),
	)
	if err != nil {
		logger.Fatal(err)
	}
	ctx := context.Background()

	defer machine.Close(ctx)
	logger.Info("Resources:")
	logger.Info(machine.ResourceNames())

	dis, err := discovery.FromRobot(machine, "discover")
	if err != nil {
		logger.Fatal(err)
	}
	cfgs, err := dis.DiscoverResources(ctx, nil)
	if err != nil {
		logger.Fatal(err)
	}
	for _, cfg := range cfgs {
		logger.Infof("Name: %v\tModel: %v\tAPI: %v", cfg.Name, cfg.Model, cfg.API)
		logger.Infof("Attributes: ", cfg.Attributes)
	}
}
```


viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,134 @@
...
+// Close closes the discovery service.
+func (dis *rtspDiscovery) Close(_ context.Context) error {
+	return nil
+}
+
+// DiscoverResources discovers different rtsp cameras that use onvif.
+func (dis *rtspDiscovery) DiscoverResources(ctx context.Context, _ map[string]any) ([]resource.Config, error) {
+	potentialCams := []resource.Config{}
+
+	list, err := viamonvif.DiscoverCameras(ctx, dis.Credentials, nil, dis.logger)
```
JohnN193, line 80: currently I am not using any manual urls. Can add these to the config if requested, pending a rescope.

viamonvif/discovery.go
```diff
@@ -109,6 +109,11 @@ type CameraInfo struct {
 	HardwareID      string   `json:"hardware_id"`
 }
 
+// Name creates generates a name for the camera based on discovered information about the camera.
+func (cam *CameraInfo) Name() string {
+	return fmt.Sprintf("%s_%s_%s", cam.Manufacturer, cam.Model, cam.SerialNumber)
+}
+
```
JohnN193, line 116: no attachment on whether this gets used as the camera name for the generated cameras. open to suggestions if people have any preferences

randhid (commented):
A few requests.

One q that isn't an ask to add anything but just curious: did you not update the readme on purpose because of the potential re-scope of the config attributes?

viamonvifdiscovery/service_test.go
```diff
@@ -0,0 +1,67 @@
...
+	cfg := Config{}
+	ctx := context.Background()
+	logger := logging.NewTestLogger(t)
+
+	t.Run("Test Default Service", func(t *testing.T) {
+		testName := "test"
+		resourceCfg := resource.Config{API: discovery.API, Model: Model, Name: testName, ConvertedAttributes: &cfg}
+		dis, err := newDiscovery(ctx, nil, resourceCfg, logger)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, dis, test.ShouldNotBeNil)
```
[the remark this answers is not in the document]
JohnN193: That we successfully create a discovery service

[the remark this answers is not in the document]
randhid: yes - but like are there values we can check in `dis` that we are are expecting?

If no it's just an empty discovery service, i'd rather do a `ShouldResemble` so we're testing something valuable about the return.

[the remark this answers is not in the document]
JohnN193: ideally I'd be able to test the DiscoverResources function, but I'm not sure the right way to mock out different cameras on a network, and I cannot import the tests in `viamonvif/discovery_test.go` into here. 

Once the project is further along and I can safely remove the DiscoverComponents implementation + move the discover service back into the `viamonvif` package, I want to look into adding tests for DiscoverResources. Until then I was planning on just leaving the test for Name().

viamonvif/discovery_test.go
```diff
@@ -64,6 +64,7 @@ func TestGetCameraInfo(t *testing.T) {
 		test.That(t, cameraInfo.SerialNumber, test.ShouldEqual, "44444444")
 		test.That(t, len(cameraInfo.RTSPURLs), test.ShouldEqual, 1)
 		test.That(t, cameraInfo.RTSPURLs[0], test.ShouldEqual, "rtsp://username:password@192.168.1.100/stream")
+		test.That(t, cameraInfo.Name(), test.ShouldEqual, "Evil Inc._Doom Ray Camera of Certain Annihilation_44444444")
```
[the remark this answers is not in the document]
JohnN193: kk added a regex to strip non alpha numerics

viamonvif/discovery.go
```diff
@@ -109,6 +110,17 @@ type CameraInfo struct {
 	HardwareID      string   `json:"hardware_id"`
 }
 
+// regex to remove non alpha numerics.
+var reg = regexp.MustCompile("[^a-zA-Z0-9 ]+")
```
randhid, line 114: Should not include spaces

meta.json
```diff
@@ -6,19 +6,33 @@
...
       "api": "rdk:component:camera",
-      "model": "viam:viamrtsp:rtsp-mjpeg"
+      "model": "viam:viamrtsp:rtsp-mjpeg",
+      "markdown_link": "README.md#configure-your-viamrtsp-camera",
+      "short_description": "configure a rtsp camera that uses the M-JPEG codec."
+    },
+    {
+      "api": "rdk:service:discovery",
+      "model": "viam:viamrtsp:onvif",
+      "markdown_link": "README.md#configure-the-viamrtsponvif-discovery-service",
```
JohnN193, line 34: saw that this was enabled on app so I added it. feels like readmes should be updated with the assumption that the markdown will be read on the config card, but can save that for a later date. 

tested the feature in staging, currently only works with registry modules. 
Screenshot 2025-01-23 at 10 24 34 PM

randhid: These are meant to link to the anchor that shows each model's specific attributes table I believe.

JohnN193: The example modules appear to include the model description, example config, attributes, and notes for the model. Easy enough to change it to be only the example config or only the attributes table though.

randhid: Nah, that's fine then , leave it as is

randhid (approved):

hexbabe:
Does this PR affect `examples/discovery-client`?

JohnN193:
> Does this PR affect `examples/discovery-client`?

it should not, made the pr in a way to not affect the current behavior

viamonvif/discovery.go
```diff
@@ -109,6 +110,17 @@ type CameraInfo struct {
...
+var reg = regexp.MustCompile("[^a-zA-Z0-9]+")
+
+// Name creates generates a name for the camera based on discovered information about the camera.
+func (cam *CameraInfo) Name() string {
+	stripManufacturer := reg.ReplaceAllString(cam.Manufacturer, "")
+	stripModel := reg.ReplaceAllString(cam.Model, "")
+	stripSerial := reg.ReplaceAllString(cam.SerialNumber, "")
+	return fmt.Sprintf("%s-%s-%s", stripManufacturer, stripModel, stripSerial)
+}
+
```
randhid, line 123: @nicksanford here is the part that you may be concerned about - this gives us the same name per model manufacturer and serial

JohnN193: I can add a number as a part of the input if people want to make the names different. 

Personally I think its okay if the same camera has the same name for different streams to help prevent users from configuring a camera multiple times.

viamonvifdiscovery/service_test.go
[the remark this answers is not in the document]
JohnN193: realized I was not testing the case for no cameras being detected, so I added an error for that and updated this test accordingly

viamonvif/discovery.go
seanavery: Makes sense to me. 

I think we are safe to assume that the name will be unique given the manufacturer and serial number are both included (pretty sure a given manufacturer guarantees unique serial numbers for their products).

viamonvif/discovery.go
seanavery: Hm.. But a user may want to configure multiple streams for a given "camera". E.g. NVR/XVR with multiple channels

JohnN193: I'll go ahead and update the function to take in a channel number if we think thats a normal use case. Other option would be to add the channel to `CameraInfo` but I'm not familiar enough with the code to know where to find that

`fmt.Sprintf("%s-%s-%s-channel%v", stripManufacturer, stripModel, stripSerial, channel)`

viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,138 @@
...
+type Config struct {
+	Credentials []device.Credentials `json:"credentials"`
+}
+
+// Validate validates the discovery service.
+func (cfg *Config) Validate(_ string) ([]string, error) {
+	// check that all creds have both usernames and passwords set. Note a credential can have both fields empty
+	for _, cred := range cfg.Credentials {
+		if cred.Pass != "" && cred.User == "" {
+			return nil, fmt.Errorf("credential missing username, has password %v", cred.Pass)
```
hexbabe, line 44: Did mobile/customer say that username only is an acceptable input?

viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Config is the config for the discovery service.
+type Config struct {
+	Credentials []device.Credentials `json:"credentials"`
+}
+
+// Validate validates the discovery service.
+func (cfg *Config) Validate(_ string) ([]string, error) {
+	// check that all creds have both usernames and passwords set. Note a credential can have both fields empty
```
hexbabe, line 41: is this comment still true

viamonvifdiscovery/service.go
JohnN193: rand mentioned it was acceptable

viamonvifdiscovery/service.go
JohnN193: naur will update

viamonvifdiscovery/service.go
hexbabe: ok I've never tried our discovery flow with just a username. Curious at what the result and output off this branch would be

viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Model is the model for a rtsp discovery service.
+var (
+	Model             = viamrtsp.Family.WithModel("onvif")
+	errNoCamerasFound = errors.New("no cameras found, ensure cameras are working or check credentials")
+)
+
+func init() {
+	resource.RegisterService(
+		discovery.API,
```
hexbabe, line 27: dumb question-- this is from RDK right

viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,138 @@
...
+	conf resource.Config,
+	logger logging.Logger,
+) (discovery.Service, error) {
+	cfg, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return nil, err
+	}
+	dis := &rtspDiscovery{
+		Named: conf.ResourceName().AsNamed(),
+		// place the empty credentials first, so we find the insecure cameras before other cameras.
```
hexbabe, line 68: what's empty about the credentials here? Aren't we just pulling in the deserialized value from the config

viamonvifdiscovery/service.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			return nil, err
+		}
+		potentialCams = append(potentialCams, camConfigs...)
+	}
+	return potentialCams, nil
+}
+
+func createCamerasFromURLs(l viamonvif.CameraInfo, logger logging.Logger) ([]resource.Config, error) {
+	potentialCams := []resource.Config{}
```
hexbabe, line 105: [nit] how do we feel about just naming this `cams`. On first read I thought there would be further filtering logic down the line, but we just return `potentialCams` as is in the outermost `DiscoverResources`

viamonvifdiscovery/service_test.go
```diff
@@ -0,0 +1,70 @@
...
+	conf, err := createCameraConfig(camName, camURL)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, conf.Name, test.ShouldEqual, camName)
+	cfg, err := resource.NativeConfig[*viamrtsp.Config](conf)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, cfg.Address, test.ShouldEqual, camURL)
+	test.That(t, *cfg.RTPPassthrough, test.ShouldBeTrue)
+}
+
+func TestDiscoveryConfig(t *testing.T) {
```
hexbabe, line 46: If we allow username only, let's add a test here

hexbabe (changes requested):
Thanks for adding me to review! Good context for me. Please see comments

viamonvifdiscovery/service.go
JohnN193: added it to rdk last release!

viamonvifdiscovery/service.go
JohnN193: oh this is another out of date comment that I missed

viamonvifdiscovery/service.go
JohnN193: I would be fine calling it cams. I only called it potentialCams because a user doesn't have to configure all of them, but yeah it does imply more filtering will happen later

hexbabe: not adamant about the suggestion, feel free to leave as is

viamonvifdiscovery/service.go
JohnN193: tested with the service, though I don't have any cameras with only a username the `DiscoverCameras` can  still return the insecure cameras when configured this way

hexbabe: gotcha

as long as the returned rtsp addresses are usable that's good

hexbabe (approved):
LGTM!

viamonvif/discovery.go
```diff
@@ -109,6 +110,17 @@ type CameraInfo struct {
...
 
+// regex to remove non alpha numerics.
+var reg = regexp.MustCompile("[^a-zA-Z0-9]+")
+
+// Name creates generates a name for the camera based on discovered information about the camera.
+func (cam *CameraInfo) Name(channel int) string {
+	stripManufacturer := reg.ReplaceAllString(cam.Manufacturer, "")
+	stripModel := reg.ReplaceAllString(cam.Model, "")
+	stripSerial := reg.ReplaceAllString(cam.SerialNumber, "")
+	return fmt.Sprintf("%s-%s-%s-channel%v", stripManufacturer, stripModel, stripSerial, channel)
```
seanavery, line 121: Wondering if we should change "channel" to "uri" or "url" to avoid conflating the index here with the actual route that may or may not include a channel number.

seanavery (approved):
LGTM!

PR #77, opened 2025-01-28
add empty cred

JohnN193:
I missed that the new discovery function still needs a credential to be passed in to run. by adding an empty credential it ensures that insecure cameras on the network are found

### DIFF
--- a/viamonvifdiscovery/service.go
+++ b/viamonvifdiscovery/service.go
@@ -20,6 +20,7 @@ import (
 var (
 	Model             = viamrtsp.Family.WithModel("onvif")
 	errNoCamerasFound = errors.New("no cameras found, ensure cameras are working or check credentials")
+	emptyCred         = device.Credentials{}
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


randhid (approved):

viamonvifdiscovery/service.go
```diff
@@ -20,6 +20,7 @@ import (
 var (
 	Model             = viamrtsp.Family.WithModel("onvif")
 	errNoCamerasFound = errors.New("no cameras found, ensure cameras are working or check credentials")
+	emptyCred         = device.Credentials{}
```
randhid, line 23: not blocking, but add test?

JohnN193: I currently cannot test this, since I need to be able to add camera mocks to these tests. 

Once the circular dependencies can be removed(after discovercomponents is deleted) we can move the discovery service into the `viamonvif` folder. then I can use the camera mocks defined in those tests and add a way to use them with onvif discovery

randhid: Right, thanks for bringing that up again.

PR #78, opened 2025-01-28
[RSDK-9833] - Add preference for TCP transport

seanavery:
## Description

Currently, the default behavior of gortsplib is to first try udp and then fallback to tcp if it is not available. This is an issue because the parser is much more stable when using TCP transport.

This is a simple PR to wrap our `reconnectClient` flow with a retry mechanism to attempt the connection using TCP, UDP, and UDP-multicast in a priority order.

!!! Due to a quirk with how Setup's are handled in gortsplib we cannot simply retry Setup calls with specified transport so we must fully teardown and restart the connection flow for each attempt.

### Manual Tests
Manual tested transport retries using mediamtx. 
We can specify available transports in the `mediamtx.yml` file and connect an appropriate fake RTSP feed using FFmpeg.
`protocols: [udp, multicast, tcp]`.
`ffmpeg -re -f lavfi -i testsrc=size=640x480:rate=30 -vcodec hevc -pix_fmt yuv420p -f rtsp -rtsp_transport <transport> rtsp://0.0.0.0:8554/live.stream`.
- udp ✅ 
- tcp ✅ 
- multicast (todo)
  - !!! even latest ffmpeg 7.1 does not support multicast.
- reconfigures ✅ 

### Automated Tests
- Added both "tcp" and "udp" transport integration tests. ✅ 
  - See succesful CI run [here](https://github.com/seanavery/viamrtsp/actions/runs/13017702730/job/36310843767).
  
  ## Profiling
  
  
Screenshot 2025-01-29 at 1 06 26 PM

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -527,11 +527,8 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 		rc.logger.Warn("no PPS found in H265 format")
 	}
 
-	res, err := rc.client.Setup(session.BaseURL, media, 0, 0)
+	_, err = rc.client.Setup(session.BaseURL, media, 0, 0)
 	if err != nil {
-		if res != nil {
-			return fmt.Errorf("status code when calling RTSP Setup on %s for H265: %w, status_code: %d", session.BaseURL, err, res.StatusCode)
-		}
 		return fmt.Errorf("when calling RTSP Setup on %s for H265: %w", session.BaseURL, err)
 	}
 
--- a/rtsp.go
+++ b/rtsp.go
@@ -235,7 +235,7 @@ func (rc *rtspCamera) clientReconnectBackgroundWorker(codecInfo videoCodec) {
 			}
 
 			if badState {
-				if err := rc.reconnectClient(codecInfo); err != nil {
+				if err := rc.reconnectClientWithTransportFallback(codecInfo); err != nil {
 					rc.logger.Warnf("cannot reconnect to rtsp server err: %s", err.Error())
 				} else {
 					rc.logger.Infof("reconnected to rtsp server url: %s", rc.u)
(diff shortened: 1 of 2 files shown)


rtsp.go
```diff
@@ -255,14 +255,44 @@ func (rc *rtspCamera) closeConnection() {
...
+		&transportUDPMulticast,
+	}
+	// Try to reconnect with each transport in the order defined above.
+	// If all attempts fail, return the last error.
+	var lastErr error
+	for _, transport := range transports {
+		if err := rc.reconnectClient(codecInfo, transport); err != nil {
+			rc.logger.Warnf("cannot reconnect to rtsp server using transport %s, err: %s", transport.String(), err.Error())
+			lastErr = err
+			continue
```
hexbabe, line 278: Should we add a small timeout here just in case we might be hitting the reconnect button too fast?

hexbabe (approved):
This LGTM as long as we comprehensively test with our office cameras and analyze the potential cons of defaulting to TCP. I'm also wondering if we need to introduce a new config attribute soon.

hexbabe:
Oh also we should remember to test reconfigures

seanavery:
@hexbabe 

> Oh also we should remember to test reconfigures
Tested out reconfigures are working ok.

> as long as we comprehensively test with our office cameras 

Tried out on a couple of cams in the office, accepting TCP transport nicely.

> I'm also wondering if we need to introduce a new config attribute soon

Me too.. I thought about adding attribute in this PR to flag pref but want to push everything possible to TCP right now since there are known issues with UDP parsing. Also UDP-multicast remains mostly untested. @randhid What do you think?

seanavery:
> Can we do a multi-hour test to see if this spikes up memory and CPU? Can be as simple as adding a CPU monitor to a machine, capturing the data from those two readings and showing the graph in teleop.

@randhid 

Added teleop results for cpu and mem. Looks to be pretty stable.

randhid (approved):

PR #80, opened 2025-01-28
[RSDK-9847] - Bump gortsplib dep

seanavery:
## Description

Bumping gortsplib to get latest improvements from h264/5 decoders. There seems to be some progress in improving fragmentation units here in the last few months.

The only deprecation change is moving from `PacketPTS` to `PacketPTS2`.

## Tests
- h264 ✅ 
  - image ✅ 
  - passthrough ✅ 
- h265 image ✅ 
- mjpeg ✅ 
- reconfigures ✅ 
- shutdown ✅


go.mod
```diff
@@ -6,22 +6,22 @@ toolchain go1.23.2
```
hexbabe, line 6: Do reconfigures/reboot behavior look ok?

rtsp_test.go
```diff
@@ -202,7 +201,7 @@ func TestRTSPCameraPerformance(t *testing.T) {
 		timeoutCtx, timeoutCancel := context.WithTimeout(context.Background(), time.Second*10)
 		defer timeoutCancel()
 		config := resource.NewEmptyConfig(camera.Named("foo"), ModelAgnostic)
-		config.ConvertedAttributes = &Config{Address: "rtsp://" + h.s.RTSPAddress}
+		config.ConvertedAttributes = &Config{Address: "rtsp://" + h.s.RTSPAddress + "/"}
```
hexbabe, line 204: Did we confirm that this doesn't double append `/`?

rtsp.go
```diff
@@ -404,12 +404,12 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
...
 
 		publishToWebRTC := func(pkt *rtp.Packet) {
-			pts, ok := rc.client.PacketPTS(media, pkt)
+			pts, ok := rc.client.PacketPTS2(media, pkt)
 			if !ok {
 				return
 			}
 			ntp := time.Now()
-			u, err := fp.ProcessRTPPacket(pkt, ntp, pts, true)
+			u, err := fp.ProcessRTPPacket(pkt, ntp, time.Duration(pts), true)
```
hexbabe, line 412: Is PTS not time since epoch?

rtsp_test.go
seanavery: Yeah I think that was a log that I inserted into gorstplib that was logging expected route and route request.

Confirmed the full RTSP URI was indeed rtsp://127.0.0.1:3251/

Moved to using `/stream1` in test cases to make things more clear (not having an empty route).

rtsp.go
seanavery: It is not necessarily time since epoch but normally time since the start of the media stream. In this case we are simply parsing out the PTS from the RTP packet we are receiving.

randhid (approved):

hexbabe (approved):

PR #81, opened 2025-01-28
[RSDK-9807]   Return RGBA bytes from viamrtsp

hexbabe:
[RSDK-9807](https://viam.atlassian.net/browse/RSDK-9807)

[RSDK-9807]: https://viam.atlassian.net/browse/RSDK-9807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Tested with my local Python SDK client
```
async def main():
    try:
        robot = await connect()
        cam = Camera.from_robot(robot, 'rtsp-1')
        count = 0
        while count < 3:
            try:
                count += 1
                img = await cam.get_image(mime_type=CameraMimeType.VIAM_RGBA)
                print(f"count: {count}")
                print(f"img.mime_type: {img.mime_type}")
                print(f"img.data: {img.data[:800]}")
                pil_img = viam_to_pil_image(img)
                pil_img.show()
            except Exception as e:
                print(f"Error getting/processing image: {e}")
                break
    finally:
        await cam.close()
        await robot.close()
```
This means that the serialized output was successfully sent over the wire and deserialized by the Python SDK, with a helper func Robin could/is using.

This PR also makes a few C/libav safety changes in mime.go with freeing the encoders, avoiding double frees, and setting jpeg enc opts

### Regression tests

- [x] h264 to yuyv 
- [x] h264 to jpeg 
- [x] h265 to yuyv 
- [x] h265 to jpeg 
- [x] live size change yuyv handling 
- [x] multiple yuyv subscribers 
- [x] `video-store`ing with `yuyv` config attr set to true

### RGBA tests

- [x] h264 to rgba
- [x] h265 to rgba
- [x] mpeg to rgba
- [x] live size change rgba handling 
- [x] multiple rgba subscribers
- [x] 2 yuyv subscriber, 2 rgba subscriber, 2 jpeg subscriber (concurrently using a Go client)

### DIFF
--- a/mime.go
+++ b/mime.go
@@ -8,13 +8,13 @@ package viamrtsp
 #include <libavutil/dict.h>
 #include <libavutil/frame.h>
 #include <libavutil/imgutils.h>
+#include <libavutil/pixfmt.h>
 #include <libswscale/swscale.h>
 #include <stdlib.h>
 */
 import "C"
 
 import (
-	"bytes"
 	"encoding/binary"
 	"errors"
 	"sync"
@@ -26,12 +26,10 @@ import (
 )
 
 const (
-	yuyvMagicString    = "YUYV"
-	yuyvHeaderDimBytes = 4
-	yuyvBytesPerPixel  = 2
-	rgbaMagicString    = "RGBA"
-	rgbaHeaderDimBytes = 8
-	rgbaBytesPerPixel  = 4
+	yuyvMagicString   = "YUYV"
+	yuyvBytesPerPixel = 2
+	rgbaMagicString   = "RGBA"
+	rgbaBytesPerPixel = 4
 )
 
 type mimeHandler struct {
@@ -130,6 +128,32 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
 	return nil
 }
 
+func (mh *mimeHandler) convertYUYV(frame *C.AVFrame) ([]byte, camera.ImageMetadata, error) {
+	return mh.convertPixelFormat(
+		frame,
+		yuyvMagicString,
+		&mh.yuyvSwsCtx,
+		&mh.yuyvFrame,
+		&mh.yuyvMu,
+		mh.initYUYVContext,
+		yuyvBytesPerPixel,
+		mimeTypeYUYV,
+	)
+}
+
+func (mh *mimeHandler) convertRGBA(frame *C.AVFrame) ([]byte, camera.ImageMetadata, error) {
+	return mh.convertPixelFormat(
+		frame,
+		rgbaMagicString,
+		&mh.rgbaSwsCtx,
+		&mh.rgbaFrame,
+		&mh.rgbaMu,
+		mh.initRGBAContext,
+		rgbaBytesPerPixel,
+		rutils.MimeTypeRawRGBA,
+	)
+}
+
 // convertPixelFormat handles the common logic for converting frames to different pixel formats
 func (mh *mimeHandler) convertPixelFormat(
 	frame *C.AVFrame,
@@ -139,7 +163,6 @@ func (mh *mimeHandler) convertPixelFormat(
 	mu *sync.Mutex,
 	initContext func(*C.AVFrame) error,
 	bytesPerPixel int,
-	headerDimBytes int,
 	mimeType string,
 ) ([]byte, camera.ImageMetadata, error) {
 	if frame == nil {
@@ -176,7 +199,7 @@ func (mh *mimeHandler) convertPixelFormat(
 	}
 
 	dataSize := int((*dstFrame).width) * int((*dstFrame).height) * bytesPerPixel
-	header := packHeader(format, int((*dstFrame).width), int((*dstFrame).height), headerDimBytes)
+	header := packHeader(format, int((*dstFrame).width), int((*dstFrame).height))
 	data := make([]byte, len(header)+dataSize)
 	copy(data[0:], header)
 	C.memcpy(unsafe.Pointer(&data[len(header)]), unsafe.Pointer((*dstFrame).data[0]), C.size_t(dataSize))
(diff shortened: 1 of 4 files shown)


mime.go
```diff
@@ -29,15 +29,22 @@ const (
...
 	jpegEnc    *C.AVCodecContext
 	yuyvFrame  *C.AVFrame
 	yuyvSwsCtx *C.struct_SwsContext
+	rgbaFrame  *C.AVFrame
+	rgbaSwsCtx *C.struct_SwsContext
 	currentPTS int
-	mu         sync.Mutex
+	jpegMu     sync.Mutex
+	yuyvMu     sync.Mutex
+	rgbaMu     sync.Mutex
```
randhid, line 47: Why the separate mutexes? Are there races we introduce and do we pass the mimehandler in by value anywhere?
https://stackoverflow.com/questions/54681678/multiple-mutexes-in-same-struct

I prefer having one mutex per struct to keep the read/write flow and locking behaviour explicit, especially since all of these are private, so we could separate them out or make nested structures.

mime.go
```diff
@@ -116,107 +122,193 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+		rgbaMagicString,
+		&mh.rgbaSwsCtx,
+		&mh.rgbaFrame,
+		&mh.rgbaMu,
+		mh.initRGBAContext,
+		rgbaBytesPerPixel,
+		rgbaHeaderDimBytes,
+		rutils.MimeTypeRawRGBA,
+	)
+}
```
randhid, line 215: can we re-organize this code so that the YUV and RGBA handlers are close to were they're first used, plus grouped together where appropriate.

randhid (changes requested):

mime.go
hexbabe: Each one protects each frame format's respective encoder/sws context. If we only used one, it would make concurrent different frame format GetImage requests slower since we're over-protecting.

hexbabe: I think the code handles the mimehandler as a pointer everywhere

hexbabe:
This PR got bigger because I decided to refactor the mime stuff since RGBA and YUYV are so similar. I need to comprehensively regression test because the scope of the implementation got less isolated.

mime.go
randhid: So if we've got three clients - a vision service, data collector and the frontend all asking for different mimetypes at the same time, what happens with the three mime mutexes?

hexbabe: IIUC this code:

The three mime mutexes are all utilized in that case. They wouldn't protect contention though because each is requesting a different mime type and there is only one subscriber per mime type.

The "intended" use of having per-mime mutexes would be if 2/3 of the subscribers all wanted one frame mimetype e.g. jpeg, but then the other 1/3 wanted a different one e.g. yuyv, then we would be properly managing contention because the yuyv can convert in parallel with the two jpeg subs do their lock contention thing.

If we had a single mutex, the yuyv would have to wait for the jpeg guys to not be currently converting, when it should just be able to do it freely.

hexbabe: Talked IRL, limiting to 1 mutex for sanity and scope

mime.go
```diff
@@ -116,107 +119,200 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+	if res := C.av_frame_get_buffer(newFrame, 32); res < 0 {
+		C.av_frame_free(&newFrame)
+		return newAvError(res, "failed to allocate buffer for frame")
 	}
-	mh.yuyvSwsCtx = C.sws_getContext(
+
+	newSwsCtx := C.sws_getContext(
 		frame.width, frame.height, C.AV_PIX_FMT_YUV420P,
-		frame.width, frame.height, C.AV_PIX_FMT_YUYV422,
+		frame.width, frame.height, pixFmt,
```
hexbabe, line 259: To summarize this PR with one line...

mime.go
```diff
@@ -116,107 +123,203 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+// for pixel format conversions. It handles cleanup of existing contexts/frames and allocation of new ones.
+//
+// Parameters:
+// - frame: Source AVFrame containing the input image
+// - pixFmt: Target pixel format to convert to
+// - swsCtxPtr: Pointer to SwsContext pointer that will be initialized
+// - dstFrame: Pointer to AVFrame pointer that will be initialized
+//
+// Returns error if any allocation or initialization fails
+func (mh *mimeHandler) initPixelFormatContext(
```
randhid, line 230: Where is this called?

randhid (approved):

seanavery (approved):
Looking good to me! Thanks for doing thorough regression testing and cleaning things up nicely.

mime.go
hexbabe: Above by the yuyv/rgba specific context helper funcs

mime.go
```diff
@@ -116,107 +123,203 @@ func (mh *mimeHandler) initJPEGEncoder(frame *C.AVFrame) error {
...
+	return mh.convertPixelFormat(
+		frame,
+		rgbaMagicString,
+		&mh.rgbaSwsCtx,
+		&mh.rgbaFrame,
+		mh.initRGBAContext,
+		rgbaBytesPerPixel,
+		rutils.MimeTypeRawRGBA,
+	)
+}
```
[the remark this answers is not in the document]
hexbabe: as args?

hexbabe:
Thanks for reviewing!

mime.go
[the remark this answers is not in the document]
randhid: Yep

PR #82, opened 2025-01-29
[RSDK-9834] - Pack NALUs in h265 decode flow

seanavery:
## Description

If we get an inter-frame (TRAIL_N or other) that is spread across multiple NALUs, which happens with large frame sizes, the decoder runs into issues since it is expecting to send and receive payloads on frame boundaries. 

This PR simply always packs the AU into a single packed payload. This ensures our decoder is able to receive a frame on the other end.

Confirmed this works for inter-frames (TRAIL_N) as well as key frame sets (VPS, SPS, PPS, IDR_W_RADL). In other words, it is ok to pack no matter what.

## Test
- steves 4k cam ✅ 
- other random office cams ✅

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -513,13 +513,18 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 		// If the AU has more than one NALU, compact them into a single payload with NALUs seperated
 		// in AnnexB format. This is necessary because the H.265 decoder expects all NALUs for a frame
 		// to be in a single payload rather than chunked across multiple decode calls.
-		var packed []byte
+		packed := []byte{}
 		for i, nalu := range au {
 			// Add start code prefix to all NALUs except the first one.
 			if i != 0 {
 				packed = append(packed, H2645StartCode()...)
 			}
 			packed = append(packed, nalu...)
+		}
+
+		if len(packed) == 0 {
+			rc.logger.Warn("no NALUs found in H265 AU skipping packet")
+			return
 		}
 
 		frame, err := rc.rawDecoder.decode(packed)
--- a/rtsp.go
+++ b/rtsp.go
@@ -514,12 +514,18 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 		// in AnnexB format. This is necessary because the H.265 decoder expects all NALUs for a frame
 		// to be in a single payload rather than chunked across multiple decode calls.
 		packed := []byte{}
-		for i, nalu := range au {
-			// Add start code prefix to all NALUs except the first one.
-			if i != 0 {
+		firstNALU := true
+		for _, nalu := range au {
+			if len(nalu) == 0 {
+				rc.logger.Warn("empty NALU found in H265 AU, skipping NALU")
+				continue
+			}
+			// Add start code prefix to all NALUs except the first non-empty one.
+			if !firstNALU {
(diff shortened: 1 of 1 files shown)


rtsp.go
```diff
@@ -510,22 +510,28 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
 			}
+			packed = append(packed, nalu...)
+		}
 
-			if frame != nil {
-				rc.handleLatestFrame(frame)
-			}
+		frame, err := rc.rawDecoder.decode(packed)
+		if err != nil {
+			// This error is created with `github.com/pkg/errors`. Explicitly call `Error()` to
```
randhid, line 527: This package isn't imported in this file?

rtsp.go
```diff
@@ -510,22 +510,28 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
-			}
-			if err != nil {
-				// This error is created with `github.com/pkg/errors`. Explicitly call `Error()` to
-				// avoid logging the stacktrace.
-				rc.logger.Debugw("error decoding(2) h265 rtsp stream", "err", err.Error())
-				return
+		// If the AU has more than one NALU, compact them into a single payload with NALUs seperated
+		// in AnnexB format. This is necessary because the H.265 decoder expects all NALUs for a frame
+		// to be in a single payload rather than chunked across multiple decode calls.
+		var packed []byte
```
randhid, line 516: This makes a `[]byte(nil)` rather than a `[]byte{}`, is this what you  want? Both lengths are 0, so please check if there's anywhere we need to to a packed byte length check.

randhid (changes requested):

rtsp.go
seanavery: Copied this comment over from previous code.. 
It is `pkg/errors` is an indirect dependency but not sure if the comment is still relevant. Can remove if confusing.

rtsp.go
```diff
@@ -510,22 +510,33 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
-			if err != nil {
-				// This error is created with `github.com/pkg/errors`. Explicitly call `Error()` to
-				// avoid logging the stacktrace.
-				rc.logger.Debugw("error decoding(2) h265 rtsp stream", "err", err.Error())
-				return
+		// If the AU has more than one NALU, compact them into a single payload with NALUs seperated
+		// in AnnexB format. This is necessary because the H.265 decoder expects all NALUs for a frame
+		// to be in a single payload rather than chunked across multiple decode calls.
+		packed := []byte{}
+		for i, nalu := range au {
```
hexbabe, line 517: Not sure how often these occur, but should we `continue` on empty NALUs?

hexbabe (approved):

rtsp.go
seanavery: If a NALU is empty... not sure how we will be hitting these situations but it would at the worst case add an extra start-code which the decoder should be able to handle.

Added check just to be safe.

rtsp.go
randhid: yeah I believe this particular file is using "errors" and the indirect dependency is coming from rdk where the cursed package is used. Let's remove this comment ) think sean yu cleaned them all out to use the standard errors package.

randhid (approved):

PR #83, opened 2025-02-04
[RSDK-9880] - Add context and timeout to ONVIF device http client

hexbabe:
[Added a unit test](https://github.com/viam-modules/viamrtsp/pull/83/commits/497c9a14bc81b1e15d66e71a1973cd4f9379397f) that [fails with the old implementation](https://github.com/viam-modules/viamrtsp/actions/runs/13143725163/job/36676791831).

[Added new logic that now requires a context, timeout, and handling for device http reqs](https://github.com/viam-modules/viamrtsp/pull/83/commits/ecdc664c0fcf01748d3ca96f80bd656215ea80df)


rtsp.go
```diff
@@ -545,7 +545,7 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
 			return
 		}
 
-		// If the AU has more than one NALU, compact them into a single payload with NALUs seperated
+		// If the AU has more than one NALU, compact them into a single payload with NALUs separated
```
hexbabe, line 548: wow the linter fixes typos now? That's crazy this change wasn't made by me

viamonvif/device/device.go
```diff
@@ -312,10 +320,14 @@ func (dev Device) callOnvifServiceMethod(endpoint string, method interface{}) ([
...
-	// this should be converted to the http interface that takes a context
-	resp, err := dev.params.HTTPClient.Post(endpoint, contentType, bytes.NewBufferString(message))
+
+	req, err := http.NewRequestWithContext(dev.ctx, http.MethodPost, endpoint, bytes.NewBufferString(message))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Content-Type", contentType)
+
+	resp, err := dev.params.HTTPClient.Do(req)
```
hexbabe, line 330: `Do` method respects ctx

viamonvif/device/device.go
```diff
@@ -88,11 +91,16 @@ type GetCapabilities struct {
 
 // NewDevice function construct a ONVIF Device entity.
 func NewDevice(params Params, logger logging.Logger) (*Device, error) {
+	if params.Context == nil {
```
nicksanford, line 94: This function should take a context and it should be required.

viamonvif/device/device_test.go
```diff
@@ -0,0 +1,79 @@
...
+		dev, err := NewDevice(Params{
+			Xaddr:      serverURL,
+			HTTPClient: &http.Client{},
+			Context:    ctx,
+		}, logger)
+		test.That(t, err, test.ShouldBeNil)
+
+		_, err = dev.sendSoap(server.URL, "test message")
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, "context deadline exceeded")
```
nicksanford, line 69: Test using test.ShouldBeError not substring matching

viamonvif/device/device_test.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+}
+
+// Helper function to read request body.
+func readBody(r *http.Request) string {
+	body, _ := io.ReadAll(r.Body)
```
nicksanford, line 75: Test for errors

viamonvif/device/device_test.go
```diff
@@ -0,0 +1,79 @@
...
+	r.Body.Close()
```
nicksanford, line 76: Can this fail? If so please assert it didn't

viamonvif/device/device_test.go
```diff
@@ -0,0 +1,79 @@
...
+	r.Body = io.NopCloser(bytes.NewBuffer(body))
```
nicksanford, line 77: Why this noopcloser?

nicksanford: Also there is a string buffer if you want to return a string

viamonvif/device/device_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/test"
+)
+
+func TestSendSoapNoHang(t *testing.T) {
```
nicksanford, line 18: Can hug please test it works when it doesn't hang?

nicksanford:
Have you run manual tests? If so, what were the tests and what were the results?

viamonvif/device/device_test.go
hexbabe: > Why this noopcloser?

It allows for the body to be read multiple times. By default, the Body of an http.Request is a `ReadCloser`, which means it can only be read once. I think we can remove it though since we only read it once in the end.

hexbabe: > Also there is a string buffer if you want to return a string
> we can remove it though since we only read it once in the end.

Done and done

viamonvif/device/device_test.go
hexbabe: Done

viamonvif/device/device_test.go
hexbabe: Done

viamonvif/device/device_test.go
hexbabe: Done. Had to cast the outer error to a url error though

viamonvif/device/device.go
hexbabe: Done

viamonvif/device/device_test.go
hexbabe: Manually tested with config

```
{
    "creds": [
        {"user": "admin", "pass": "admin"}
    ],
    "xaddrs": ["http://127.0.0.1:8080"]
  }
```

And Python script that pretends to be an ONVIF http client, then hangs on subsequent requests:

```
from http.server import BaseHTTPRequestHandler, HTTPServer
import time
import xml.etree.ElementTree as ET

class HangHandler(BaseHTTPRequestHandler):
    def do_POST(self):
        # Read the request body
        content_length = int(self.headers.get('Content-Length', 0))
        body = self.rfile.read(content_length).decode('utf-8')
        
        # Parse the SOAP request
        try:
            root = ET.fromstring(body)
            # Find all elements in the SOAP body using namespace-aware search
            for elem in root.iter():
                if 'GetCapabilities' in elem.tag:
                    print("Received GetCapabilities request. Responding...")
                    self.send_response(200)
                    self.send_header('Content-Type', 'application/soap+xml')
                    self.end_headers()
                    response = '''<?xml version="1.0" encoding="UTF-8"?>
                        <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://www.w3.org/2003/05/soap-envelope">
                            <SOAP-ENV:Body>
                                <GetCapabilitiesResponse>
                                    <Capabilities>
                                        <Analytics>
                                            <XAddr>http://127.0.0.1:8080/onvif/analytics</XAddr>
                                        </Analytics>
                                        <Device>
                                            <XAddr>http://127.0.0.1:8080/onvif/device</XAddr>
                                        </Device>
                                        <Events>
                                            <XAddr>http://127.0.0.1:8080/onvif/events</XAddr>
                                        </Events>
                                        <Imaging>
                                            <XAddr>http://127.0.0.1:8080/onvif/imaging</XAddr>
                                        </Imaging>
                                        <Media>
                                            <XAddr>http://127.0.0.1:8080/onvif/media</XAddr>
                                        </Media>
                                        <PTZ>
                                            <XAddr>http://127.0.0.1:8080/onvif/ptz</XAddr>
                                        </PTZ>
                                    </Capabilities>
                                </GetCapabilitiesResponse>
                            </SOAP-ENV:Body>
                        </SOAP-ENV:Envelope>'''
                    self.wfile.write(response.encode())
                    return
                elif 'GetDeviceInformation' in elem.tag:
                    self.send_response(200)
                    self.send_header('Content-Type', 'application/soap+xml')
                    self.end_headers()
                    response = '''<?xml version="1.0" encoding="UTF-8"?>
                        <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://www.w3.org/2003/05/soap-envelope">
                            <SOAP-ENV:Body>
                                <GetDeviceInformationResponse>
                                    <Manufacturer>Python Test Device</Manufacturer>
                                    <Model>Hang Simulator</Model>
                                    <FirmwareVersion>1.0</FirmwareVersion>
                                    <SerialNumber>123456</SerialNumber>
                                    <HardwareId>HW123</HardwareId>
                                </GetDeviceInformationResponse>
                            </SOAP-ENV:Body>
                        </SOAP-ENV:Envelope>'''
                    self.wfile.write(response.encode())
                    return
        except ET.ParseError:
            pass
            
        # For any other request, simulate a hang
        print("Received request that will hang. Request body:", body)
        time.sleep(1000)

def run(server_class=HTTPServer, handler_class=HangHandler, port=8080):
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    print(f'Starting httpd server on port {port}')
    httpd.serve_forever()

if __name__ == "__main__":
    run()
```
that runs on the IP+port defined in the config.

Running `bin/darwin-arm64/discovery -config ./config.json`, we see the log `2025-02-05T18:50:33.486Z        WARN    discovery       viamonvif/discovery.go:157      Failed to get camera info from http://127.0.0.1:8080: failed to get RTSP URLs: failed to get media profiles: Post "http://127.0.0.1:8080/onvif/media": context deadline exceeded` and the process cleanly shuts down with no hang.

cmd/discovery/cmd.go
```diff
@@ -11,12 +11,15 @@ import (
...
 	"os"
 	"slices"
+	"time"
 
 	"github.com/viam-modules/viamrtsp/viamonvif"
 	"github.com/viam-modules/viamrtsp/viamonvif/device"
 	"go.viam.com/rdk/logging"
 )
 
+const timeoutDuration = 10 * time.Second
```
seanavery, line 21: Curious why 10? Seems pretty long. Is this to take into account WS Discovery across the network interfaces?

hexbabe: I just chose 10 to bias for having a longer timeout + it felt right. Bad reason haha, but since the cmd is a solely dev tool, I figured it'd be ok to make that decision

seanavery (commented):
Making sense that we did not have timeouts on the Onvif HTTP calls prior which could lead to hangs.

One thing I am confused about is whether or not we need a separate context handler specifically for HTTP calls like `callOnvifServiceMethod`.

Currently is this passing a general 10 second timeout to `DiscoverCameras` which is passed through to all subsequent calls? Would it make sense to add another handler for 100ms to each Onvif call?

hexbabe:
> Making sense that we did not have timeouts on the Onvif HTTP calls prior which could lead to hangs.
> 
> One thing I am confused about is whether or not we need a separate context handler specifically for HTTP calls like `callOnvifServiceMethod`.
> 
> Currently is this passing a general 10 second timeout to `DiscoverCameras` which is passed through to all subsequent calls? Would it make sense to add another ctx timeout for 100ms to each Onvif call?

I'm not too worried about the 10 second timeout for the cmd tool since it's only used by us and Eliot for debugging. In "real" viam-server scenarios, we'll use the context given to use by the Go SDK module server, which should have a cancellation/signal handler.

cmd/discovery/cmd.go
randhid: Let's make this an argument input to the test script. And leave it at this default value default

cmd/discovery/cmd.go
```diff
@@ -38,7 +41,9 @@ func main() {
 }
 
 func realMain() error {
-	ctx := context.Background()
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), timeoutDuration)
+	defer cancel()
```
randhid, line 45: Log something when we timeout please

viamonvif/device/device.go
```diff
@@ -307,15 +308,20 @@ func (dev Device) callOnvifServiceMethod(endpoint string, method interface{}) ([
...
-	resp, err := dev.params.HTTPClient.Post(endpoint, contentType, bytes.NewBufferString(message))
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewBufferString(message))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Content-Type", contentType)
+
+	// Using Do instead of POST to support context cancellation and timeout.
+	resp, err := dev.params.HTTPClient.Do(req)
```
randhid, line 324: My understanding i sthat the fix is in these lines - by changing the interface to the one that can take in a context and not hang.
Can you run the new discovery binary against the olde code and verify that we're not missing any of the discovered devices in office?

viamonvif/discovery.go
```diff
@@ -142,7 +142,7 @@ func DiscoverCameraInfo(
 			return zero, fmt.Errorf("context canceled while connecting to ONVIF device: %s", xaddr)
 		}
 		// This calls GetCapabilities
-		dev, err := device.NewDevice(device.Params{
+		dev, err := device.NewDevice(ctx, device.Params{
```
randhid, line 145: Do we want to pass in the client's context? Okay for now - but it can come with a timeout of its own. Which is why I want a context cancelled log where appropriate to ensure we know that discovery timed out because of a Deadlien or context cancellation. You can try this out by running a client script against the Discover Resources API and verifying that you get useful information on what happened.

hexbabe: > client's context

IIUC, if we run `viamrtsp` as a module, this `ctx` is a module server provided context passed in from the `Discover` registered func, which is passed into `DiscoverCameras`, which in turn is used in the invoked function `DiscoverCameraInfo` containing the quoted line of code.

When running the discovery cmd debug tool, we bypass the `Discover` registry and directly make the call `viamonvif.DiscoverCameras(timeoutCtx, opts.config.Creds, urls, logger)`. `timeoutCtx` will be the `ctx` in question within the quoted `NewDevice` instantiation.

These are two different code paths, one with a cancellation/possible future timeout from the module server, and another with the timeout we specify in the `cmd/discovery/cmd.go`.

What I just said may be wrong, but if I am right then I don't think we need to worry about conflating the module server passed context and cmd's `timeoutCtx`.

cmd/discovery/cmd.go
hexbabe: Added an error log

hexbabe:
Sorry for the messy history and diff now. Should've rebased instead of merging in main

viamonvif/discovery.go
randhid: No - I mean do you want the client'scall's context here or is it better to control that context ourselves for a default timeout? 

For motors actuating methods for example - we have a separate context for some routines to ensure they're polling polling position so we have full control of that context in the driver.  Passing in the context from the client should be fine in that case - I am curious what the behaviour is if you pass in a small timeout in a script using this API and if that behaviour clearly reports that a timeout happened?

hexbabe: Wait when you say client do you mean the debugging binary compiled from `cmd/discovery`

randhid: Nope - sdk code calling the discovery service from robot and the proper DiscoverResources API.

hexbabe: Gotcha

viamonvif/device/device.go
hexbabe: Ran the discovery example client on main vs. feature with some known credentials and both gave the same exact outputs

viamonvif/discovery.go
hexbabe: I set a tiny timeout
```
	timeoutCtx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
	defer cancel()
	cfgs, err := dis.DiscoverResources(timeoutCtx, extras
```
And I observed that it basically immediately timed out with
`2025-02-26T18:30:13.565Z        ERROR   client  discovery-client/client.go:50   Error occurred: context canceled`

hexbabe: Also verified that the binary tool hits timeouts
```seanyu@ROBOT-G26PRFHC2H sean-viamrtsp % /Users/seanyu/Projects/sean-viamrtsp/bin/darwin-arm64/discovery -timeout 1 
2025-02-26 13:34:17.706 discovery[39524:1310139] WARNING: AVCaptureDeviceTypeExternal is deprecated for Continuity Cameras. Please use AVCaptureDeviceTypeContinuityCamera and add NSCameraUseContinuityCameraDeviceType to your Info.plist.
2025-02-26T18:34:19.723Z        WARN    discovery       viamonvif/discovery.go:49       failed to connect to ONVIF device context canceled while connecting to ONVIF device: 192.168.1.1
2025-02-26T18:34:19.723Z        ERROR   discovery       discovery/cmd.go:73     Discovery timed out after 1s```

seanavery (approved):

PR #85, opened 2025-02-05
[RSDK-9920] - add lazy_decode

nicksanford:
### Changes:

https://viam.atlassian.net/browse/RSDK-9920

1. Add new config parameter: `lazy_decode: bool`, default: false
2. Only applies if codec is h264. If set and codec is anything else, log a warning at startup.
3. If set to true, accumulate the NALUs (h264 messages) since the last IDR. Only compute an image from those NALUs when camera.Image is called.

### Performance Impact:

**Hardware: Pi 5:**

> Resolution: 2304*1296 (larger than 1080p)
> Frame Rate (FPS): 25
> Max Bitrate (Kbps): 4096
> I-frame Interval: 2x (about every 2 seconds)
> 

**Baseline `lazy_decode: false` or unset:** 
1. no callers: htop reports ~35% cpu 
2. with gostream calling camera.Image as fast as it can:  htop reports ~50% cpu, each request takes ~35 ms

**`lazy_decode: true`**
1. no callers: htop reports ~2% cpu 
2. with gostream calling camera.Image as fast as it can:  htop reports ~50% cpu, each request takes 100-500 ms, depending on how frequently camera.Image is called. When called more frequently, fewer NALUs need to be processed and the image can be computed and returned faster. 

Here is a video of the experience of each config option.
1. first is with rtp_passthrough: true (best and the default)
2. second is with  rtp_passthrough: false, lazy_decode: false
3. third is with  rtp_passthrough: false, lazy_decode: true

https://github.com/user-attachments/assets/975d3d0d-442d-4a3f-ba5c-456b54861c74

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -384,7 +384,7 @@ func (rc *rtspCamera) consumeAU() {
 	defer rc.auMu.Unlock()
 	if len(rc.au) > 0 {
 		rc.storeH264Frame(rc.au)
-		rc.au = [][]byte{}
+		rc.au = nil
 	}
 }
 
--- a/rtsp.go
+++ b/rtsp.go
@@ -163,6 +163,8 @@ type rtspCamera struct {
 	gostream.VideoReader
 	u *base.URL
 
+	auMu       sync.Mutex
+	au         [][]byte
 	client     *gortsplib.Client
 	rawDecoder *decoder
 
@@ -370,6 +372,25 @@ func (rc *rtspCamera) reconnectClient(codecInfo videoCodec, transport *gortsplib
 	return nil
 }
 
+func (rc *rtspCamera) consumeAU() {
+	rc.auMu.Lock()
+	defer rc.auMu.Unlock()
+	rc.storeH264Frame(rc.au)
+	rc.au = [][]byte{}
+}
+
+func (rc *rtspCamera) resetAU(au [][]byte) {
+	rc.auMu.Lock()
+	defer rc.auMu.Unlock()
+	rc.au = au
+}
+
+func (rc *rtspCamera) appendAU(au [][]byte) {
+	rc.auMu.Lock()
+	defer rc.auMu.Unlock()
+	rc.au = append(rc.au, au...)
+}
+
 // initH264 initializes the H264 decoder and sets up the client to receive H264 packets.
 func (rc *rtspCamera) initH264(session *description.Session) (err error) {
 	// setup RTP/H264 -> H264 decoder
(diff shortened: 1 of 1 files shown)


rtsp.go
```diff
@@ -161,8 +162,13 @@ type rtspCamera struct {
 	resource.AlwaysRebuild
 	model resource.Model
 	gostream.VideoReader
-	u *base.URL
+	u             *base.URL
+	lazyDecodeRTP bool
 
+	closeMu sync.RWMutex
```
nicksanford, line 168: I needed to add this as now both Image and OnPacket's callback (which are called from 2 different goroutines) can try to interact with the pool and the decoder. As a result, I made it so that `Image` can execute concurrently with other `Image` calls but can't run concurrently with `Close` and such that it will early exist if close has already been called.

rtsp.go
```diff
@@ -106,6 +106,7 @@ func init() {
 type Config struct {
 	Address        string               `json:"rtsp_address"`
 	RTPPassthrough *bool                `json:"rtp_passthrough"`
+	LazyDecode     bool                 `json:"lazy_decode"`
```
randhid, line 109:
```suggestion
	LazyDecode     bool                 `json:"lazy_decode,omitempty"`
```

rtsp.go
randhid: This sounds like it should be implemented with our `StoppableWorker` structure to handle this exact case, can you make a TODO and a ticket to look into this. 

Are you seeing cases where Close is not stopping the Image call? If so there's probably a bug or something being badly freed/uncancelled in this code.

rtsp.go
```diff
@@ -250,6 +258,7 @@ func (rc *rtspCamera) closeConnection() {
 		rc.client.Close()
 		rc.client = nil
 	}
+	rc.resetAU(nil)
```
randhid, line 261: [q] Why reset to nil rather than an empty array of bytes like in the consumer helper?

randhid (approved):

rtsp.go
nicksanford: why is this needed? I think omitempty only changes the behavior if you are marshalling to json, which this struct will never do.

rtsp.go
nicksanford: good question. I'll change `consumeAU` to set rc.au to `nil`

rtsp.go
nicksanford: It is a similar shape of problem to what we use `StoppableWorker` to solve, however I don't think `StoppableWorker` is the right solution to this problem as the goroutine calling `Image` is spawned by the camera/server.go hosted in the module, not by our code.

I think the approach I used here is reasonable. Close won't do it's thing until all in progress `Image` calls complete and once `Close` is called, `Image` won't execute.

rtsp.go
randhid: Good q - eventually app/fleet want to consume these config structs to show fancy configs, so I put them in for the far flung future.

PR #87, opened 2025-02-06
Raw segmenter

seanavery:
## Description

This PR allows for an efficient video storage path in the viamrtsp module by passing through encoded RTSP feed directly into a segmenter.

> !!! Relies on changes in video-store repo. See PR [here](https://github.com/viam-modules/video-store/pull/37).

## Todo:
- [x] Working build with video-store as a library
- [x] Get playable video from h264 feed 
- [x] Split clips by keyframe 
- [ ] Add config attribute for video storage
- [ ] Get playable video from h265 feed
- [ ] Add hooks for do_command interface
  - Figure out how this plays with concater and deleter functionality inside of video-store

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -504,7 +504,8 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
 			packed = append(packed, lengthBytes...)
 			packed = append(packed, nalu...)
 		}
-		err = rc.rawSegmenter.WritePacket(packed, int64pts)
+		isIDR := h264.IDRPresent(au)
+		err = rc.rawSegmenter.WritePacket(packed, int64pts, isIDR)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


rtsp.go
```diff
@@ -390,6 +398,13 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
 		return fmt.Errorf("creating H264 RTP decoder: %w", err)
 	}
 
+	// Segmenter needs its own rtp decoder since the parser it is stateful
+	// and will not work concurrently with the parser used for the decode path.
+	rtpSegDec, err := f.CreateDecoder()
```
seanavery, line 403: We should be able to share the same gorstplib decoder (parser) between store and segment paths.

nicksanford:
moved here: https://github.com/viam-modules/viamrtsp/pull/93

PR #95, opened 2025-02-11
[RSDK-9965] discover resources using extra

JohnN193:
https://viam.atlassian.net/browse/RSDK-9965

can't add a test for this yet until after DiscoverComponents removal. 

tested with a go client. will be testing with mobile team

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ The following attributes are available for all models of `viamrtsp` discovery se
 | ------- | ------ | ------------ | ----------- |
 | `credentials` | struct | Optional | set the username and password for any amount of credentials. |
 
-### Example configuration
+### Example Configuration
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -11,14 +11,14 @@ This module also implements the `"rdk:service:discovery"` API, to surface rtsp c
 * `viam:viamrtsp:onvif` - discovers cameras using the [onvif interface](https://www.onvif.org/).
 
 
-## Configure your `viamrtsp` camera
-
 Navigate to the [**CONFIGURE** tab](https://docs.viam.com/build/configure/) of your [machine](https://docs.viam.com/fleet/machines/) in the [Viam app](https://app.viam.com/).
 [Add the camera component to your machine](https://docs.viam.com/build/configure/#components), searching for `viamrtsp` and selecting your desired model.
 
+## Configure your `viamrtsp` camera
+
 Copy and paste the following attributes template into the resulting component's attribute panel:
 
-```
+```json
 {
    "rtp_passthrough": true,
    "rtsp_address": "rtsp://foo:bar@192.168.10.10:554/stream"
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -91,11 +91,22 @@ The following attributes are available for all models of `viamrtsp` discovery se
     {
       "user": "<USERNAME2>",
       "pass": "<PASSWORD2>"
-    },
```
randhid, line 94: While we're here - let's move this to the top and run viam module update.

viamonvifdiscovery/service.go
```diff
@@ -74,10 +74,22 @@ func newDiscovery(_ context.Context, _ resource.Dependencies,
...
+	if ok {
+		// not requiring a password to match config
+		extraPass, ok := extra["Pass"].(string)
+		if !ok {
+			extraPass = ""
+		}
+		addCred := device.Credentials{User: extraUser, Pass: extraPass}
+		discoverCreds = append(discoverCreds, addCred)
+	}
+	list, err := viamonvif.DiscoverCameras(ctx, discoverCreds, nil, dis.logger)
```
randhid, line 92: If extra is nil I think the user check will panic. 
Let's:
1. Make this a helper function to return the creds to pass to viamonvif.DiscoverCamers.
2. Bail if extra is nil
3. Write a test to ensure you're getting what you want out of the reutns from the new function.

randhid (changes requested):

README.md
JohnN193: sure, ill get the other fixes for this doc too then

viamonvifdiscovery/service.go
JohnN193: nil didn't have an issue when testing in client code, are there other ways it could happen?

JohnN193: also I can't do the tests I want until I can mock cameras, which won't be until after we move into the `viamonvif` folder

JohnN193: Can add the helper + test it tho

randhid: Add the tst - I wonder if we're passing through an an empty map[string]interface{} at this point to be defensive.
But no harm in being extra defensive

README.md
```diff
@@ -11,14 +11,14 @@ This module also implements the `"rdk:service:discovery"` API, to surface rtsp c
...
 
 
-## Configure your `viamrtsp` camera
-
 Navigate to the [**CONFIGURE** tab](https://docs.viam.com/build/configure/) of your [machine](https://docs.viam.com/fleet/machines/) in the [Viam app](https://app.viam.com/).
 [Add the camera component to your machine](https://docs.viam.com/build/configure/#components), searching for `viamrtsp` and selecting your desired model.
 
+## Configure your `viamrtsp` camera
+
 Copy and paste the following attributes template into the resulting component's attribute panel:
```
randhid, line 20:
```suggestion

Copy and paste the following attributes template into the JSON configuration:

```

viamonvifdiscovery/service_test.go
```diff
@@ -68,3 +68,42 @@ func TestDiscoveryConfig(t *testing.T) {
...
+	})
+	t.Run("Test good extra with no Pass", func(t *testing.T) {
+		extra := map[string]any{
+			"User": "user",
+		}
+		cred, ok := getCredFromExtra(extra)
+		test.That(t, cred.User, test.ShouldEqual, "user")
+		test.That(t, cred.Pass, test.ShouldEqual, "")
+		test.That(t, ok, test.ShouldBeTrue)
+	})
```
randhid, line 101:
```suggestion
	t.Run("Test good extra with no Pass", func(t *testing.T) {
		extra := map[string]any{
			"User": "user",
		}
		cred, ok := getCredFromExtra(extra)
		test.That(t, cred.User, test.ShouldEqual, "user")
		test.That(t, cred.Pass, test.ShouldEqual, "")
		test.That(t, ok, test.ShouldBeTrue)
	})
		t.Run("Test bad extra with nostrings" func(t *testing.T) {
		extra := map[string]any{
			"User": 1,
			"Pass" : true,
		}
		cred, ok := getCredFromExtra(extra)
		test.That(t, cred.User, test.ShouldEqual, "")
		test.That(t, cred.Pass, test.ShouldEqual, "")
		test.That(t, ok, test.ShouldNotBeTrue)
	})
```

randhid (approved):
Few nits, please add that test case but good to go!

PR #102, opened 2025-02-26
RSDK-9953: Have discovery create mDNS mappings. Return URLs leveraging the mapping.

dgottlieb:


viamonvif/discovery.go
```diff
@@ -163,6 +200,26 @@ func DiscoverCameraInfo(
 	return zero, fmt.Errorf("no credentials matched IP %s", xaddr)
 }
 
+var consecutiveDashesRegexp = regexp.MustCompile(`\-\-+`)
+
+func strToHostName(inp string) string {
```
nicksanford, line 205: could you add a comment with a link to the spec / standard this is comfirming to?

viamonvif/discovery.go
```diff
@@ -121,6 +125,39 @@ func (cam *CameraInfo) Name(urlNum int) string {
...
+
+	// Replace the URLs in-place such that configs generated from these objects will point to the
+	// logical dns hostname rather than a raw IP.
+	for idx := range cam.RTSPURLs {
+		if strings.Contains(cam.RTSPURLs[idx], cam.deviceIP.String()) {
+			cam.RTSPURLs[idx] = strings.Replace(cam.RTSPURLs[idx], cam.deviceIP.String(), derivedHostname, 1)
+		} else {
+			logger.Debugf("RTSP URL did not contain expected hostname. URL: %v HostName: %v",
+				cam.RTSPURLs[idx], cam.deviceIP.String())
+			mdnsServer.Remove(cleanedSerialNumber)
```
nicksanford, line 156: I've seen cases (such as in the AV checkin with eliot this week) where a camera returns urls with multiple hostnames (such as 192.168.x.x and 166.x.x.x). If that happens, I would expect we would want to still add the mdns entry for 192.168.x.x  even if there are other garbage ips in the RTSPURLs list.

Should we maybe only call remove if NONE of the `cam.RTSPURLs` contained the deviceID?

viamonvif/mdns_server.go
```diff
@@ -0,0 +1,93 @@
...
+		if info.ip.Equal(ip) {
+			// We're already managing an mdns entry for the same hostname -> ip mapping.
+			return nil
+		}
+
+		info.cleanup()
+	}
+	delete(server.mappedDevices, serialNumber)
+
+	avoidLintMagicNumberComplaintThatIsntEvenRelevant := 8080
```
nicksanford, line 46: cute

viamonvif/mdns_server.go
```diff
@@ -0,0 +1,93 @@
...
+	mdnsServer, err := zeroconf.RegisterProxy(
+		serialNumber, // Dan: As far as I can tell, just a name.
+		"_rtsp._tcp", // Dan: The mDNS "service" to register. Doesn't make a difference?
```
nicksanford, line 49: It is possible for the communication between viamrtsp and the camera to be over udp.  Will that cause problems with this `.tcp` name?

viamonvif/mdns_server.go
nicksanford: sorry, why do we need this?

viamonvif/discovery_test.go
```diff
@@ -131,3 +135,55 @@ func TestExtractXAddrsFromProbeMatch(t *testing.T) {
...
+	test.That(t, strToHostName("A______3"), test.ShouldEqual, "A-3")
+	test.That(t, strToHostName("--__--A--__--3--__--"), test.ShouldEqual, "-A-3-")
+	test.That(t, strToHostName("$$__$$A$$__$$3$$__$$"), test.ShouldEqual, "-A-3-")
+}
+
+func TestMDNSMapping(t *testing.T) {
+	// This test relies on having access to multicast network interfaces. Cloud instances often
+	// disable multicast capabilities. We skip this test by default as we can't expect CI testing to
+	// succeed.
+	t.Skip("fails in CI")
```
nicksanford, line 154: why does this fail in ci?

viamonvif/discovery.go
dgottlieb: > I've seen cases (such as in the AV checkin with eliot this week) where a camera returns urls with multiple hostnames (such as 192.168.x.x and 166.x.x.x).

Interesting. Can you share a fuller onvif response/state of `CamInfo`?

> Should we maybe only call remove if NONE of the cam.RTSPURLs contained the deviceID?

That seems like a reasonable approach.

viamonvif/discovery_test.go
nicksanford: when I run:

cd viamonvif && go test -v -race -run TestMDNSMapping -count=10 -failfast

I see errors in eric's mdns library but I don't see test failures.

I see that his shutdown is swallowing any errors: https://github.com/edaniels/zeroconf/blob/master/server.go#L195C4-L195C12

Is this cause for concern or a red herring?

```
=== RUN   TestMDNSMapping
    discovery_test.go:170: 2025-02-26T10:31:09.333-0500 INFO    TestMDNSMapping viamonvif/discovery_test.go:170 Conjured DNS Name:jMxllMahxb.local
    logger.go:146: 2025-02-26T10:31:14.617-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:835  mdns: failed to set multicast interface {"error": "raw-control udp4 0.0.0.0:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.618-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:854  mdns: failed to set multicast interface {"error": "raw-control udp6 [::]:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.618-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:835  mdns: failed to set multicast interface {"error": "raw-control udp4 0.0.0.0:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.618-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:854  mdns: failed to set multicast interface {"error": "raw-control udp6 [::]:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.619-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:835  mdns: failed to set multicast interface {"error": "raw-control udp4 0.0.0.0:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.619-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:854  mdns: failed to set multicast interface {"error": "raw-control udp6 [::]:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.619-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:835  mdns: failed to set multicast interface {"error": "raw-control udp4 0.0.0.0:5353: use of closed network connection"}
    logger.go:146: 2025-02-26T10:31:14.619-0500 DEBUG   TestMDNSMapping zeroconf@v1.0.10/server.go:854  mdns: failed to set multicast interface {"error": "raw-control udp6 [::]:5353: use of closed network connection"}
--- PASS: TestMDNSMapping (5.29s)
```

viamonvif/discovery.go
dgottlieb: Are you asking about the spec that defines what makes a valid DNS name?

Or [this line](https://docs.google.com/document/d/1bxObh_p7Ix94gEhmvwnPDW23TzJp21NptaadMRqXG5A/edit?tab=t.0#bookmark=id.rrxrkcoj02ig) that describes we want the property where the same input will continue to map to the same output?

The intention of the immediately following documentation was to copy the necessary context over. I'm happy to add a blurb about predictability and more explicitly state this is all of the context behind this algorithmic decision.

viamonvif/mdns_server.go
nicksanford: like... why do we need to specify a port? What does this port mean?

dgottlieb: I have to pass a number into this `RegisterProxy` API call. Even though, as far as I can tell, it has no impact.

But when I just call this as:
```
	mdnsServer, err := zeroconf.RegisterProxy(
		serialNumber, // Dan: As far as I can tell, just a name.
		"_rtsp._tcp", // Dan: The mDNS "service" to register. Doesn't make a difference?
		"local",      // the domain
		8080,         // The service's port is ignored here
		serialNumber,          // actual mDNS hostname, without the .local domain
		[]string{ip.String()}, // ip to use
		[]string{},            // txt fields, not needed
		nil,                   // resolve this name for requests from all network interfaces
		// RSDK-8205: logger.Desugar().Sugar() is necessary to massage a ZapCompatibleLogger into a
		// *zap.SugaredLogger to match zeroconf function signatures.
		server.logger.Desugar().Sugar(),
	)
```

the linter complains about using a magic number. So the API requires a number, but the number being chosen can be anything. It is the _opposite_ of magic.

I don't know if this nonsense is specific to zeroconf, or if it's actually part forwarding expectations from the system call API.

nicksanford:
I ran a manual test in the office and it worked.

viamonvif/mdns_server.go
dgottlieb: I didn't see any problems in testing regarding tcp/udp. Strictly speaking, ping is _neither_ tcp nor udp. Nor is ping seeking the rtsp service. And ping does* work/stop working as this RegisterProxy/"mdnsServer.Shutdown" methods are called.

This is admittedly pretty magic. Thus the documentation on every input of what I found to be necessary and how it maps to the side-effects a user can see. I'm happy to run a specific experiment.

More broadly, there's buy-in from product that if this just doesn't seem to be working, we'll try a different approach such as editing /etc/hosts. Or using "macros" or "templates" such that the viam-server does the IP substitution before passing a configuration to camera construction/reconfigure.

viamonvif/discovery_test.go
dgottlieb: > why does this fail in ci?

dgottlieb: > I see that his shutdown is swallowing any errors: https://github.com/edaniels/zeroconf/blob/master/server.go#L195C4-L195C12

> Is this cause for concern or a red herring?

In my experience, returning errors when closing resources is an anti-pattern. What is the user supposed to do? Retry?

So I'm not particularly worried that the underlying library is masking that. Happy to have the future prove me wrong and we can document a specific scenario that needs to be guarded against.

viamonvif/discovery.go
dgottlieb: I can confirm that the `xaddr.Host` IP we're using to substitute is the IP that we sent the [onvif discovery request to](https://github.com/viam-modules/viamrtsp/blob/ae313acae6a26e03df70676d8b533bb05cc0d1a6/viamonvif/discovery.go#L45-L47). So this code can additionally go wrong if an onvif response comes from a device that never advertises in any RTSPUrl the IP it handled a request from.

We can alleviate that with defensive code to parse every RTSP url and do a substitution, but now we'd need multiple dns names (I don't think it's useful to map a single dns name to multiple IPs if only some of them are going to work -- that'd be the state we're in right now).

At the risk of adding a bunch of hard to follow defensive code, I think the logging we have will inform us if this we're hitting a case where multiple IPs in different RTSP urls is being problematic.

viamonvif/discovery_test.go
dgottlieb: > Is this cause for concern or a red herring?

Oh sorry, I missed actually reading the error messages. I don't believe those are coming from Close, but rather from register itself? 

I'm not excited to see those messages, but given those are all udp 0.0.0.0 ips, I'm not particularly worried. It's good to know though that debug mode will print out some zeroconf information even if* it returns a successful result.

viamonvif/discovery.go
nicksanford: valid DNS name.

nicksanford (approved):
looks good to me

JohnN193:
Running into an issue while testing. 

If the viam-server gets restarted without a call to the discovery service, a rtsp camera using the mdns mapping will fail. Not too sure what the fix would be

JohnN193:
you could make cameras discovered this way have an implicit dependency on the discovery service. that way if an rtsp camera starting fails to find the camera it can re ping the discovery service to setup mdns again

dgottlieb:
> If the viam-server gets restarted without a call to the discovery service, a rtsp camera using the mdns mapping will fail. Not too sure what the fix would be

That's intended behavior for this ticket. https://viam.atlassian.net/browse/RSDK-10101 is addressing that which I have mostly done.

JohnN193 (approved):
ah didn't see the 2nd ticket, LGTM then!

JohnN193:
forgot to mention, I'm currently working on adding a discovery service for [upnp](https://github.com/viam-modules/viamrtsp/pull/103). Guessing we would want mDNS mappings for those as well so I can try to mess around with that once this is merged

dgottlieb:
> ah didn't see the 2nd ticket, LGTM then!

Sorry, was a ninja edit!

PR #103, opened 2025-02-26
[RSDK-9840] add UPnP discovery service

JohnN193:
https://viam.atlassian.net/browse/RSDK-9840

"hardware test" using the Sonos speakers: 

config: 
Screenshot 2025-02-26 at 6 40 04 PM
output: 
Screenshot 2025-02-26 at 6 39 54 PM


upnpdiscovery/service.go
```diff
@@ -0,0 +1,230 @@
...
+	cams := []resource.Config{}
+
+	discoverQueries := dis.queries
+
+	extraQuery, ok := getQueryFromExtra(extra)
+	if ok {
+		discoverQueries = append(discoverQueries, extraQuery)
+	}
+	// worth noting that the discovered hosts are not guaranteed to be rtsp camera.
+	// We assume that the user knew what they were looking for.
```
JohnN193, line 124: this is based on the behavior eliot used in UPNP_DISCOVER previously.

meta.json
```diff
@@ -34,6 +34,12 @@
       "markdown_link": "README.md#configure-your-viamrtsp-camera", 
       "short_description": "configure a rtsp camera that uses the MPEG4 codec."
     },
+    {
+      "api": "rdk:service:discovery",
+      "model": "viam:viamrtsp:onvif",
```
randhid, line 39:
```suggestion
      "model": "viam:viamrtsp:upnp",
```

rtsp.go
```diff
@@ -145,16 +145,16 @@ func (conf *Config) parseAndFixAddress(ctx context.Context, logger logging.Logge
 	}
 
 	if u.Hostname() == "UPNP_DISCOVER" {
-		host, err := viamupnp.FindHost(ctx, logger, conf.Query)
+		hosts, _, err := viamupnp.FindHost(ctx, logger, []viamupnp.DeviceQuery{conf.Query}, false)
```
randhid, line 148: add a comment with a TODO to remove this from the main rtsp logic/config.

upnpdiscovery/service.go
```diff
@@ -0,0 +1,229 @@
+// Package upnpdiscovery provides the discovery service that wraps UPnP integration for the viamrtsp module
+package upnpdiscovery
```
randhid, line 2: Can you make a cmd/upnpdiscovery script to run, that has been useful for onvif discovery on machine.

upnpdiscovery/service.go
```diff
@@ -0,0 +1,229 @@
...
+	// convert to map to be used as attributes in resource.Config
+	if err = json.Unmarshal(jsonBytes, &result); err != nil {
+		return resource.Config{}, err
+	}
+
+	return resource.Config{
+		Name: name, API: camera.API, Model: viamrtsp.ModelAgnostic,
+		Attributes: result, ConvertedAttributes: &attributes,
+	}, nil
+}
```
randhid, line 198: can make this common to discovery services - we should move it to the main rtsp package.

upnpdiscovery/service.go
```diff
@@ -0,0 +1,229 @@
...
+		camName = fmt.Sprintf("%s-%s", camName, stripManufacturer)
+	}
+	if stripModel := reg.ReplaceAllString(query.ModelName, ""); stripModel != "" {
+		camName = fmt.Sprintf("%s-%s", camName, stripModel)
+	}
+	if stripSerial := reg.ReplaceAllString(query.SerialNumber, ""); stripSerial != "" {
+		camName = fmt.Sprintf("%s-%s", camName, stripSerial)
+	}
+	return camName
+}
```
randhid, line 175: Is this the same as the onvif discover service's?

randhid (changes requested):
Wrong registered model name is the biggest thing, I would like to dry up the code a little too.

upnpdiscovery/service.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/erh/viamupnp"
+	"github.com/viam-modules/viamrtsp"
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/rdk/logging"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/discovery"
+)
+
+// Model is the model for a rtsp discovery service.
```
randhid, line 19:
```suggestion
// Model is the model for a upnp discovery service.
```

rtsp.go
```diff
@@ -145,16 +145,16 @@ func (conf *Config) parseAndFixAddress(ctx context.Context, logger logging.Logge
...
 		if err != nil {
 			return nil, err
 		}
 
 		p := u.Port()
 		if p == "" {
-			u.Host = host
+			u.Host = hosts[0]
```
seanavery, line 155: Are we ok to hardcode that the first host in the list returned from `viamupnp.FindHost` is the one we want? Or maybe its not that big of deal because we will eventually get rid of the `UPNP_DISCOVER` special case host?

seanavery:
@JohnN193 Is the upnp (or ssdp) discovery process only able to fetch host IP addresses and not available ports?

rtsp.go
JohnN193: yeah ideally we get rid of UPNP_DISCOVER down the line. though there might be some world that we dont and instead lean into using UPNP_DISCOVER, but thats a future question

JohnN193:
> @JohnN193 Is the upnp (or ssdp) discovery process only able to fetch host IP addresses and not available ports?

yes, currently we only fetch IP addresses. I wasn't able to test enough with cameras to see if there was a way to pass the port/endpoint thru discovery, and the testing I did with the sonos implied that devices have multiple ports, some of which you may not want to use. 

In the future we might have the setup/data to return ports, but for now I thought it was fine to simply match what the current discovery behavior was

upnpdiscovery/service.go
JohnN193: nope, onvif doesn't explicitly call out which endpoint is which. 

the naming is also different since we are less likely to have all of the model/serial/manufacturer info. In the future we could try getting upnp discovery to match onvif, but that would require more changes in eliots repo.

upnpdiscovery/service.go
JohnN193: the behavior is different on this technically since I wanted to return the query in the attributes as well to inform users how the device was found.

Can remove the query and make this one function if requested.

randhid (approved):

seanavery (approved):

PR #104, opened 2025-02-26
RSDK-10101: Add a cache of mdns entries for fast restart.

dgottlieb:


viamonvif/mdns_server.go
```diff
@@ -16,14 +21,29 @@ type mdnsInfo struct {
 type mdnsServer struct {
 	mu            sync.Mutex
 	mappedDevices map[string]mdnsInfo
+	cacheFilepath string
 	logger        logging.Logger
 }
 
-func newMDNSServer(logger logging.Logger) mdnsServer {
-	return mdnsServer{
+func newMDNSServer(logger logging.Logger) *mdnsServer {
```
dgottlieb, line 28: The change to a pointer was because sync.Mutex objects cannot be copied. I knew when I wrote this in the prior PR I was being too cute leaving `mdnsServer`s return value as a value type. And one patch later it bit me.

JohnN193 (commented):
so this wouldn't cover the case if the discovery service was removed from the machine, and then the machine gets restarted. 

for GOST stuff I don't think they plan on removing the discovery service from machines so I don't know if we care about the "hidden" dependency for other users.

dgottlieb:
> so this wouldn't cover the case if the discovery service was removed from the machine, and then the machine gets restarted.

Correct. We could move that more into the module half -- not entirely opposed to that. But it's certainly the case that if discovery is removed and IPs changed, there's no way to get into a good spot other than identifying IPs by hand and hard coding.

dgottlieb:
> so I don't know if we care about the "hidden" dependency for other users.

As far as how I analyze that, if a user hits this problem, they have a workaround available that doesn't require a patch+release. And we'll also have their feedback about why they wanted to remove the discovery service. And can make a more informed choice about what to do.

viamonvif/mdns_server.go
```diff
@@ -82,6 +102,89 @@ func (server *mdnsServer) Remove(hostname string) {
...
+
+	return results
+}
+
+func (server *mdnsServer) apply(newEntries []cachedEntry) {
+	// For each entry (presumably from the `cachedFilepath`), call `mdnsServer.Add`.
+	for _, entry := range newEntries {
+		ip := net.ParseIP(entry.IP)
+		if ip == nil {
+			server.logger.Warn("Unknown IP from cache file. DNSName:", entry.DNSName, "IP:", entry.IP)
```
JohnN193, line 142: [q] worth running `DiscoverResources` in the background when the resource is configured as well? so if an ip address gets changed there will be less downtime for the user.

JohnN193: Though I guess if we did that we wouldn't really need to have the cache file.

JohnN193 (changes requested):

viamonvif/mdns_server.go
dgottlieb: I think a cache file largely eliminates the need, yes. But I do want to explore what you're describing with https://viam.atlassian.net/browse/RSDK-10102

viamonvif/service.go
```diff
@@ -63,13 +66,23 @@ func newDiscovery(_ context.Context, _ resource.Dependencies,
...
 	}
 
+	// viam-server sets this environment variable. The contents of this directory is expected to
+	// persist across process restarts and module upgrades.
+	moduleDataDir := os.Getenv("VIAM_MODULE_DATA")
+	if !strings.HasPrefix(moduleDataDir, "/") {
+		dis.mdnsServer = newMDNSServer(logger)
+	} else {
+		dis.mdnsServer = newMDNSServerFromCachedData(
+			filepath.Join(moduleDataDir, "mdns_cache.json"), logger.Sublogger("mdns"))
```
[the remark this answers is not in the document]
dgottlieb: Sorry, struggled parsing this. Adding some formatting to make sure I understand.

> can we have a check in rtsp.go when rtsp cameras fail, to:
> - check if this file exists, 
> - and if it does that they should re-enable the discovery service? 
>
> rather than just having cameras fail with no such host

Can I do this in https://viam.atlassian.net/browse/RSDK-10102 ? I'm about to start working on it. To me this falls into the "viamrtsp cameras know there's discovery code in the same program". Which informed how I scoped out these tickets.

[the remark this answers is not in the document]
JohnN193: oh yeah that sentence makes no sense sorry about that. I meant have the error log tell the user `re-enable the discovery service` instead of saying `no such host`

makes sense to think about this in the next ticket as well. I misunderstood what was happening between the tickets you linked yesterday.

viamonvif/mdns_server.go
```diff
@@ -82,6 +102,89 @@ func (server *mdnsServer) Remove(hostname string) {
...
+	DNSName string `json:"dnsname"`
+	IP      string `json:"ip"`
+}
+
+func (server *mdnsServer) readCacheFile() []cachedEntry {
+	// If the `cacheFilepath` was given, read the contents and parse it as json. On any error we log
+	// and return nil, which is interpreted as an empty slice.
+	if server.cacheFilepath == "" {
+		return nil
+	}
```
JohnN193, line 115: [note to self] cacheFilepath is not given when the discovery service cannot find `VIAM_MODULE_DATA`

JohnN193 (approved):

PR #105, opened 2025-02-26
add-h265-embeded-video-store-support

nicksanford:
Add support to capture video in h265.

This does introduce the possible failure mode of https://viam.atlassian.net/browse/RSDK-10130 which will be addressed as a follow up.

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -184,13 +184,14 @@ type rtspCamera struct {
 	lazyDecode       bool
 	iframeOnlyDecode bool
 
-	closeMu    sync.RWMutex
-	vsConfig   *videostore.Config
-	vs         videostore.RTPVideoStore
-	auMu       sync.Mutex
-	au         [][]byte
-	client     *gortsplib.Client
-	rawDecoder *decoder
+	closeMu  sync.RWMutex
+	vsConfig *videostore.Config
+	// vs         videostore.RTPVideoStore
+	videoStoreMuxer *videoStoreMuxer
+	auMu            sync.Mutex
+	au              [][]byte
+	client          *gortsplib.Client
+	rawDecoder      *decoder
 
 	cancelCtx  context.Context
 	cancelFunc context.CancelFunc
@@ -286,13 +287,13 @@ func (rc *rtspCamera) closeConnection() {
 		rc.rawDecoder = nil
 	}
 
-	if rc.vs != nil {
+	if rc.videoStoreMuxer != nil {
 		ctx, cancel := context.WithTimeout(context.Background(), videoStoreInitCloseTimeout)
 		defer cancel()
-		if err := rc.vs.Close(ctx); err != nil {
+		if err := rc.videoStoreMuxer.Close(ctx); err != nil {
 			rc.logger.Errorf("videostore.Close error: %s", err.Error())
 		}
-		rc.vs = nil
+		rc.videoStoreMuxer = nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


rtsp.go
```diff
@@ -494,16 +484,20 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
...
+		vs, err := videostore.NewRTPVideoStore(config, rc.logger)
+		if err != nil {
+			return err
+		}
+		rc.videoStoreMuxer = &videoStoreMuxer{
+			SourceType: config.Type,
+			vs:         vs,
+			sps:        f.SPS,
+			pps:        f.PPS,
+			logger:     rc.logger,
```
nicksanford, line 500: This change makes it so that h264 and h265 video-stores don't depend on the initial sps, pps (and in the caes of h265 vps sent in the sdp). If the camera sends these only in the rtsp stream, video-store will still work.

We should update the storeImage code path to also be resillient to those parameters not being sent in the SDP as a follow up.

rtsp.go
```diff
@@ -677,46 +630,54 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
+		}
+		rc.videoStoreMuxer = &videoStoreMuxer{
+			SourceType: config.Type,
+			vs:         vs,
+			vps:        f.VPS,
+			sps:        f.SPS,
+			pps:        f.PPS,
+			logger:     rc.logger,
 		}
+		if err := rc.videoStoreMuxer.initVideoStore(); err == nil {
```
nicksanford, line 672: don't need this.

rtsp.go
```diff
@@ -184,13 +184,13 @@ type rtspCamera struct {
...
-	closeMu    sync.RWMutex
-	vsConfig   *videostore.Config
-	vs         videostore.H264RTPVideoStore
-	auMu       sync.Mutex
-	au         [][]byte
-	client     *gortsplib.Client
-	rawDecoder *decoder
+	closeMu         sync.RWMutex
+	vsConfig        *videostore.Config
+	videoStoreMuxer *videoStoreMuxer
```
nicksanford, line 189: I swapped out the videostore with a videoStoreMuxer 

This wraps the videostore and allows simplifying the videostore API

rtsp.go
```diff
@@ -619,19 +564,26 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
...
-		onPacketRTP = func(pkt *rtp.Packet) {
-			publishToWebRTC(pkt)
-			au, err := rtpDec.Decode(pkt)
-			if err != nil {
-				if !errors.Is(err, rtph264.ErrNonStartingPacketAndNoPrevious) && !errors.Is(err, rtph264.ErrMorePacketsNeeded) {
-					rc.logger.Debugf("error decoding(1) h264 rstp stream %w", err)
-				}
-				return
+	onPacketRTP := func(pkt *rtp.Packet) {
+		pts, ok := rc.client.PacketPTS2(media, pkt)
```
nicksanford, line 570: I slightly changed the structure to first try to get the PTS (which is needed for all other operations) and then pass the result to the functions that care about it.

rtsp.go
```diff
@@ -1567,3 +1547,318 @@ func toFetchCommand(command map[string]interface{}) (*videostore.FetchRequest, e
...
+		filteredAU = append(filteredAU, nalu)
+	}
+
+	au = filteredAU
+
+	if au == nil {
+		return
+	}
+
+	if err := e.maybeReInitVideoStore(); err != nil {
```
seanavery, line 1703: [nit] Should we only call this if we get an SPS NALU?

rtsp.go
```diff
@@ -727,7 +678,36 @@ func (rc *rtspCamera) initH265(session *description.Session) (err error) {
...
+		// Extract access units from RTP packets
+		au, err := rtpDec.Decode(pkt)
+		if err != nil {
+			if !errors.Is(err, rtph265.ErrNonStartingPacketAndNoPrevious) && !errors.Is(err, rtph265.ErrMorePacketsNeeded) {
+				rc.logger.Debugw("error decoding(1) h265 rstp stream", "err", err.Error())
+			}
+			return
+		}
+		storeImage(au)
+		rc.videoStoreMuxer.writeH265(au, pts)
```
seanavery, line 701: [q] Still a little unclear about what the behavior of the rawsegmenter in the case that first AU write here is not a keyframe.

nicksanford: we won't call it without a keyframe

rtsp.go
seanavery: Nvm! The `maybeReInitVideoStore` is no-op unless we get a SPS Nalu.

seanavery (approved):
LGTM. Happy with manual tests.

PR #106, opened 2025-02-27
RSDK-10102: Have camera connection errors run discovery.

dgottlieb:


viamonvif/service.go
```diff
@@ -18,6 +19,23 @@ import (
...
+var someDiscoveryServiceObj atomic.Pointer[rtspDiscovery]
+var someLoggerObj atomic.Pointer[logging.Logger]
+
+// poke looks to see if there is a discovery service running. If so, ask it to run discovery and
+// create any new mdns mappings. If there is no discovery service and* there happens to be an mdns
+// mapping cache file from a prior run, log a warning saying that discovery should be re-enabled.
+func poke() {
+	svc := someDiscoveryServiceObj.Load()
+	if svc == nil {
+		(*someLoggerObj.Load()).Warn("No discovery service")
```
dgottlieb, line 31: Logger is an interface which makes this wonky -- can ignore for the draft

viamonvif/service.go
```diff
@@ -136,11 +136,20 @@ func (dis *rtspDiscovery) Close(_ context.Context) error {
 	return nil
 }
 
-func createCamerasFromURLs(l CameraInfo, logger logging.Logger) ([]resource.Config, error) {
+func createCamerasFromURLs(l CameraInfo, discoveryDependencyName string, logger logging.Logger) ([]resource.Config, error) {
```
dgottlieb, line 139: This threading of the discovery dependency and which urls/configs are gaining a discovery dependency for config generation feels a bit clumsy. The root of why* it mostly has to look this way is because CameraInfo is used for [json serialization](https://github.com/viam-modules/viamrtsp/blob/0f7334dfc7a544aaa9038fc32cbd44b985adfa7f/viamonvif/discovery.go?plain=1#L107). And I didn't feel comfortable proposing a change to that user facing API.

JohnN193 (approved):
LGTM! I actually really liked how the discovered rtsp camera starts getting validation errors when the discovery service gets removed. makes it pretty clear to the user whats happening

viamonvif/service.go
JohnN193: would it be worth making this function a method on the discovery service instead? then we could 

- check the mapped devices directly
- grab the name from the service rather than passing it in

prob makes it alot harder to test and this works as is, so probably not worth the effort.

dgottlieb: > would it be worth making this function a method on the discovery service instead? then we could

Agreed, that's a reasonable direction.

> prob makes it alot harder to test and this works as is, so probably not worth the effort.

I'll take your offer to leave it as-is.

PR #107, opened 2025-02-28
[RSDK-10009] - Skip TLS verification for local IP addresses

hexbabe:
[RSDK-10009](https://viam.atlassian.net/browse/RSDK-10009)

### Manual Testing (TODO)

### DIFF
--- a/viamonvif/device/device.go
+++ b/viamonvif/device/device.go
@@ -8,10 +8,12 @@ package device
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/xml"
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -62,6 +64,9 @@ type Params struct {
 	Username   string
 	Password   string
 	HTTPClient *http.Client
+	// SkipLocalTLSVerification controls whether TLS certificate verification is skipped for local IP addresses.
+	// This is necessary for cameras with self-signed certificates.
+	SkipLocalTLSVerification bool
 }
 
 // GetProfiles is a request to the GetProfiles onvif endpoint.
@@ -87,6 +92,49 @@ type GetCapabilities struct {
 	Category onvif.CapabilityCategory `xml:"tds:Category"`
 }
 
+func isLocalIPAddress(hostname string) bool {
+	hostname = strings.TrimSpace(hostname)
+
+	if hostname == "localhost" {
+		return true
+	}
+
+	ip := net.ParseIP(hostname)
+	if ip == nil {
+		return false
+	}
+
+	if ip.IsLoopback() {
+		return true
+	}
+
+	// IPv4 private addresses.
+	if ip4 := ip.To4(); ip4 != nil {
+		// 10 prefixed addresses e.g. 10.0.0.0/8.
+		if ip4[0] == 10 { //nolint:mnd
+			return true
+		}
+		// 172.16.0.0/12.
+		if ip4[0] == 172 && ip4[1] >= 16 && ip4[1] <= 31 {
+			return true
+		}
+		// 192.168.0.0/16.
+		if ip4[0] == 192 && ip4[1] == 168 {
+			return true
+		}
+	}
+
+	// IPv6 unique local addresses (fc00::/7).
+	if ip.To16() != nil && ip.To4() == nil {
+		// ULA addresses start with fc or fd.
+		if ip[0]&0xfe == 0xfc { //nolint:mnd
+			return true
+		}
+	}
+
+	return false
+}
+
 // NewDevice function construct a ONVIF Device entity.
 func NewDevice(ctx context.Context, params Params, logger logging.Logger) (*Device, error) {
 	dev := &Device{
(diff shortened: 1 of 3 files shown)


viamonvif/device/device.go
```diff
@@ -87,6 +92,49 @@ type GetCapabilities struct {
...
+	ip := net.ParseIP(hostname)
+	if ip == nil {
+		return false
+	}
+
+	if ip.IsLoopback() {
+		return true
+	}
+
+	// IPv4 private addresses.
```
dgottlieb, line 111: I think you just want [this method](https://pkg.go.dev/net/netip#Addr.IsPrivate) from the Go standard library

hexbabe: Thanks

viamonvif/device/device_test.go
```diff
@@ -91,3 +92,106 @@ func readBody(r *http.Request) (string, error) {
...
 	return buf.String(), nil
 }
+
+func TestTLSVerificationConfig(t *testing.T) {
+	logger := logging.NewTestLogger(t)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, err := readBody(r)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
```
randhid, line 102: why not have a test.That here explicitly?

viamonvif/device/device_test.go
```diff
@@ -91,3 +92,106 @@ func readBody(r *http.Request) (string, error) {
...
+								<Media>
+									<XAddr>http://example.com/onvif/media</XAddr>
+								</Media>
+							</Capabilities>
+						</GetCapabilitiesResponse>
+					</SOAP-ENV:Body>
+				</SOAP-ENV:Envelope>`))
+		}
+	}))
+	defer server.Close()
```
randhid, line 121: We should also test that the close does not error.. You can defer that entire test case.

viamonvif/device/device_test.go
```diff
@@ -91,3 +92,106 @@ func readBody(r *http.Request) (string, error) {
 	}
 	return buf.String(), nil
 }
+
+func TestTLSVerificationConfig(t *testing.T) {
+	logger := logging.NewTestLogger(t)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
seanavery, line 99: Probably more effort than it is worth, but wondering if we could create a test http server here that uses self-signed TLS cert.

It does seem sufficient to just test that we are skipping verification when dealing with a "private" address.

seanavery (approved):
Looking good to me!

Just to clarify my understanding of the problem that this addressing -- some cameras support https using a self-signed certificate and not from a certificate authority (CA)? And the golang http client was failing on verification since it is self-signed?

viamonvif/device/device_test.go
```diff
@@ -91,3 +92,106 @@ func readBody(r *http.Request) (string, error) {
...
+			isLocal:                  true,
+			skipLocalTLSVerification: false,
+			expectSkipVerify:         false,
+		},
+		{
+			name:                     "IP public, skip enabled",
+			isLocal:                  false,
+			skipLocalTLSVerification: true,
+			expectSkipVerify:         false,
+		},
```
randhid, line 146: Do we want this case at all?

In my mind - the case that is bad user experience is a local IP address which we are being overly secure with on a person's private network. But this test case seems slightly too permissive to me.

dgottlieb (approved):

viamonvif/device/device.go
```diff
@@ -97,7 +102,27 @@ func NewDevice(ctx context.Context, params Params, logger logging.Logger) (*Devi
 	}
 
 	if dev.params.HTTPClient == nil {
-		dev.params.HTTPClient = new(http.Client)
+		var skipVerify bool
+		if params.SkipLocalTLSVerification {
```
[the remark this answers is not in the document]
hexbabe: Right, I followed the  treat viamonvif as a library  pattern since it used to be a third-party lib

Would you prefer to treat it as a direct part of the codebase for code additions?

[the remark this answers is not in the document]
randhid: Ah I see. No that makes sense then, feel free to resolve.

viamonvif/device/device_test.go
hexbabe: Oh wait I think server.Close() doesn't return error

viamonvif/device/device_test.go
hexbabe: Nope changed

viamonvif/device/device.go
```diff
@@ -97,7 +102,27 @@ func NewDevice(ctx context.Context, params Params, logger logging.Logger) (*Devi
...
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: skipVerify, //nolint:gosec
+			},
+		}
+		dev.params.HTTPClient = &http.Client{
+			Transport: transport,
+		}
+
+		if skipVerify {
+			logger.Warnf("TLS certificate verification disabled for local IP address: %s.",
```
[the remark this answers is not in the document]
hexbabe: Make it info

viamonvif/device/device_test.go
hexbabe: Good call. Not too difficult to add. Added

viamonvif/device/device_test.go
randhid: Can you walk me through the expected behaviour of the public IPs only - I want to make sure we're testing the correct things.

hexbabe: For "IP public, skip enabled," this test should pass without any errors, confirming that the device can connect to the public IP **without** skipping TLS verification. Basically making sure that the http client isn't bypassing the verification for it, therefore asserting `expectSkipVerify:         false`

For "IP public, skip disabled," this is the most trivial case where we assert that we aren't skipping TLS verification even when the flag is disabled.

viamonvif/device/device_test.go
randhid: Blergh, okay.

randhid (approved):

PR #108, opened 2025-03-04
[RSDK-10071] - Fix avFramePool memory leak

seanavery:
## Description

There is an issue where we are never releasing the `latestFrame` cache back to the pool during closeout. When the avFramePool closes out the latest frame never gets freed causing a lingering frame allocation.

### Manual Testing
- We can see that when properly releasing the frame we are now freeing more frames during closeout. See the following logs -- first log set is from upstream and the second is the feature branch.
- Manually verified memory usage with `top` does not increase from disabling and re-enabling viamrtsp component.

```
3/3/2025, 4:52:34 PM info rdk:component:camera/rtsp-cam-3     host/pool.go:112 Num pool frames remaining at close: 1   log_ts 2025-03-03T21:52:34.154Z

```

```
3/3/2025, 4:57:51 PM info rdk:component:camera/rtsp-cam-3     host/pool.go:112 Num pool frames remaining at close: 2   log_ts 2025-03-03T21:57:51.675Z

```

### Automated Tests

- Added test case to make sure we put as many frames as received back into the pool.
  - If you comment out the close frame put we see the test fails with a lingering frame.
  
 ```
      --- FAIL: TestRTSPCamera/H264 (5.84s)
        --- FAIL: TestRTSPCamera/H264/AvFramePool (3.80s)
            rtsp_test.go:122: Expected: '16'
                Actual:   '15'
                (Should be equal)
FAIL
FAIL    github.com/viam-modules/viamrtsp        5.943s
FAIL
```

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -232,6 +232,16 @@ func (rc *rtspCamera) Close(_ context.Context) error {
 	rc.unsubscribeAll()
 	rc.activeBackgroundWorkers.Wait()
 	rc.closeConnection()
+	// Clean up latest frame if it exists. This is necessary to ensure that the frame is properly
+	// freed when the avFramePool is closed.
+	rc.latestFrameMu.Lock()
+	if rc.latestFrame != nil {
+		if refCount := rc.latestFrame.decrementRefs(); refCount == 0 {
+			rc.avFramePool.put(rc.latestFrame)
+		}
+		rc.latestFrame = nil
+	}
+	rc.latestFrameMu.Unlock()
 	rc.avFramePool.close()
 	rc.mimeHandler.close()
 	return nil
--- a/rtsp.go
+++ b/rtsp.go
@@ -232,7 +232,7 @@ func (rc *rtspCamera) Close(_ context.Context) error {
 	rc.unsubscribeAll()
 	rc.activeBackgroundWorkers.Wait()
 	rc.closeConnection()
-	// Clean up latest frame if it exists. This is necessary to ensure that the frame is properly
+	// Clean up latestFrame cache if it exists. This is necessary to ensure that the frame is properly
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by hexbabe — APPROVED** [review]
Nice! Tricky

File: rtsp.go [hunk 1]
```diff
@@ -232,8 +232,18 @@ func (rc *rtspCamera) Close(_ context.Context) error {
 	rc.unsubscribeAll()
 	rc.activeBackgroundWorkers.Wait()
 	rc.closeConnection()
-	rc.avFramePool.close()
 	rc.mimeHandler.close()
+	// Clean up latestFrame cache if it exists. This is necessary to ensure that the frame is properly
+	// freed when the avFramePool is closed.
+	rc.latestFrameMu.Lock()
```
randhid, line 238: one q - we're already in a lock; closeMu it is a RWMutex, but latestFrameMu is a Mutex. Why was the extra lock necessary to get the latest frame.

Do we need to free the frame cache at other parts of the code?

randhid (approved):

rtsp.go
seanavery: Good question! I thought about removing this as well but concerned about the following:
`Close` is triggered and we  `closeConnection` which cancels the context on `gortsplib.Client. It seems like we should be safe to not have to lock since we will no longer have `handleLatestFrame` hits downstream of the `c.client.OnPacketRTP` callback. But, there still could be a race condition where we close the client and move on to the rest of Close but it takes some time for the client to go silent.

The only other concern is conflicts with the `Image` method but it has a `closeMu` read lock on it so seem to be safe there.

randhid: Okay - keep an eye on this behaviour as you battle test.

PR #112, opened 2025-03-10
[RSDK-10002] - make video store viamrtsp model

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-10002)

Depends on: https://github.com/viam-modules/video-store/pull/54

Changes:
1. Add `viam:viamrtsp:video-store` model as a generic component which implements the same `DoCommand` the video-store model in the video-store module does. That resource doesn't take a `sync` config param. Config is defined here: https://github.com/viam-modules/viamrtsp/pull/112/files#diff-b151fd77787feb836a509df656993f17d30f3ae6300dc39d96c7b6e8098403a4R12. It does allow the camera to be unspecified and if so it will still boot and allow the user to query video via `DoCommand(fetch)` and `DoCommand(save)`.
2. If the camera is specified but doesn't support the codecs video-store supports (h264 and h265 in this PR) it will return an error from the video-store contstructor and resource graph will retry.
3. If the camera started off using a supported codec and then it changed (either via reconfiguration or b/c the IP camera started sending video in a different codec) video-store will emit a log ever 5 seconds stating it is waiting for video data until it gets video data. This should be an infrequent condition.
5. Delete the embedded video-store and make DoCommand unimplemented
6. Create global registry in registry.go to allow viamrtsp cameras to register themselves for video-stores to find, and for viamrtsp cameras to send video to the video-stores that request it.
7. Add a Mux struct that remembers the sps, pps, and for h265 vps during the video stream as well as decodes dts and normalizes dts and pts to 0 before passing it to the raw segmenter. This allows video-store to start collecting video mid rtsp stream and prevents video-store from needing to force viamrtsp to restart the video in order to produce uncorrepted video segment files.
8. Add videoRequest struct to manage the lifecycle of the videostore's request to viamrtsp for video.
 
Deps changes:
1. Upgrade RDK from v0.57.1-0.20250108163155-2071720cd8d3 to v0.65.0
10. Upgrade goutils from v0.1.118 to v0.1.130
11. Upgrade viam api from v0.1.380  to v0.1.396
12. Upgrade video-store from v0.0.5-rc2 -  v0.0.5-rc4

```json
{
  "agent": {
    "version_control": {
      "viam-server": "0.65.0"
    }
  },
  "components": [
    {
      "name": "viamrtsp-video-store",
      "api": "rdk:component:generic",
      "model": "viam:viamrtsp:video-store",
      "attributes": {
        "camera": "cam",
        "storage": {
          "size_gb": 1
        }
      }
    },
    {
      "name": "cam",
      "model": "viam:viamrtsp:rtsp",
      "attributes": {
        "rtsp_address": "...",
        "lazy_decode": true
      }
    }
  ],
  "modules": [
    {
      "type": "local",
      "name": "viamrtsp-module-1",
      "executable_path": "..."
    }
  ]
}
```

Nick Manual Testing:
1. Changing `storage.size_gb`, `storage.upload_path` and `storage.storage_path`
2. Removing `camera` and confirming that video can still be queried
3. Configuring `camera` to a non viamrtsp camera (logs errors until fixed)
4. Configuring `camera` to a non existant resource name
5. Configuring `camera` to a resource that is not a camera
6. Disabling viamrtsp camera videostore is depending on (logs errors until fixed)
7. Disabling video-store while viamrtsp is receiving video from it
8. Configuring 2 video stores to consume video from the same viamrtsp camera (doesn't work for second video-store and logs errors until fixed)
9. Changing the IP camera viamrtsp is consuming from from a video-store supported codec (such has h264 or h265) to an unsupported codec (such as mpeg4), logs errors until fixed
10. Starting videostore when viamrtsp is using an unsupported codec (logs errors until fixed)
11. Configuring multiple video-stores and viamrtsp cameras


cmd/remote/cmd.go
```diff
@@ -46,7 +46,7 @@ func mainWithArgs(ctx context.Context, _ []string, logger logging.Logger) error
 		},
 	}
 
-	myRobot, err := robotimpl.New(ctx, conf, logger)
```
nicksanford, line 49: signature of this function changed in https://github.com/viamrobotics/rdk/pull/4782/files

videorequest.go
```diff
@@ -0,0 +1,104 @@
+package viamrtsp
```
nicksanford, line 1: this managest the lifecycle of the videostore's request for video

seanavery:
@nicksanford Would you mind providing example config to test this out?

nicksanford:
> @nicksanford Would you mind providing example config to test this out?

added to the pr descriptions

videostore/mux.go
```diff
@@ -0,0 +1,559 @@
...
+		logger:  logger,
+	}
+}
+
+// init and close are called by videostore.
+func (m *rawSegmenterMux) init() error {
+	if m.rawSeg == nil {
+		return errors.New("videostore.RTPVideoStore.Segmenter() is nil")
+	}
+	cam, err := registry.Global.Get(m.camName.String())
```
seanavery, line 69: Trying to understand the flow better here. Is there any possibility of race condition in`registry.Global` reference between vaimrtrtsp and video store spin up? Is it guaranteed that the camera will be registered before the muxer attempts to reference it here?

videostore/mux.go
```diff
@@ -0,0 +1,559 @@
...
+		m.logger.Errorf("AnnexBMarshal err: %s", err.Error())
+		return err
+	}
+
+	if !m.metadata.firstTimeStampsSet {
+		m.metadata.firstPTS = pts
+		m.metadata.firstDTS = dts
+		m.metadata.firstTimeStampsSet = true
+	}
+	err = m.rawSeg.WritePacket(packed, pts-m.metadata.firstPTS, dts-m.metadata.firstDTS, idrPresent)
```
seanavery, line 504: Mind adding a comment about why we need to adjust timestamps?

videostore/mux.go
```diff
@@ -0,0 +1,559 @@
...
+
+func (m *rawSegmenterMux) close() error {
+	if m == nil {
+		return nil
+	}
+	m.worker.Stop()
+	return nil
+}
+
+func (m *rawSegmenterMux) Start(codec videostore.CodecType, au [][]byte) error {
```
seanavery, line 148: Do wee need a separate `Start` method here even though we handle metadata updates in `Write`?

videostore/videostore.go
```diff
@@ -0,0 +1,216 @@
...
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/logging"
+	"go.viam.com/rdk/resource"
+)
+
+const (
+	defaultSegmentSeconds      = 30 // seconds
+	defaultUploadPath          = ".viam/capture/video-upload"
+	defaultStoragePath         = ".viam/video-storage-viamrtsp"
```
seanavery, line 20: Should we delete the `defaultStoragePath` and `maxGRPCSize` consts in rtsp.go file now?

nicksanford: good call, done

videostore/mux.go
nicksanford: yep done

videostore/mux.go
nicksanford: I don't think so.

If the video-store's config has a camera then the video-store's config will return that camera as a dependency of video-store. Resource graph will not boot video-store until the camera is booted.

seanavery:
@nicksanford Is it ok to have multiple `viam:viamrtsp:video-store` model components pointed at the same camera?

nicksanford:
> > @nicksanford Is it ok to have multiple `viam:viamrtsp:video-store` model components pointed at the same camera?
> 
> It is not allowed, but it is a condition I guard against.

Meaning it shouldn't break anything but the second video-store wont succeed in configuring and will log errors until one or the other video-store are removed.

seanavery (approved):

videostore/config.go
```diff
@@ -0,0 +1,71 @@
...
+	Camera  *string `json:"camera,omitempty"`
+	Storage Storage `json:"storage"`
+}
+
+// Storage is the storage subconfig for videostore.
+type Storage struct {
+	SizeGB      int    `json:"size_gb"`
+	UploadPath  string `json:"upload_path,omitempty"`
+	StoragePath string `json:"storage_path,omitempty"`
+}
```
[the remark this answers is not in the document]
nicksanford: correct. viam:viamrtsp:video-store has a subset of the config as the video-store module's model

videostore/mux.go
```diff
@@ -0,0 +1,573 @@
...
+	if err := m.cam.CancelRequest(m); err != nil {
+		m.logger.Warnf("DeRegister video-store from viamrtsp camera %s", err.Error())
+	}
+}
+
+func (m *rawSegmenterMux) close() error {
+	if m == nil {
+		return nil
+	}
+	m.worker.Stop()
```
randhid, line 144: I don't know how we'd get into this state in this particular - but usually we also check that the worker is nil before stopping it.

videostore/mux.go
```diff
@@ -0,0 +1,573 @@
...
+			if len(nalu) == 0 {
+				continue
+			}
+			//nolint:mnd
+			typ := h264.NALUType(nalu[0] & 0x1F)
+			switch typ {
+			case h264.NALUTypeSPS:
+				m.metadata.sps = nalu
+			case h264.NALUTypePPS:
+				m.metadata.pps = nalu
```
randhid, line 166: TODO: write what these codecs are

randhid (approved):

PR #113, opened 2025-03-12
RSDK-10243: Revert usage of mdns addresses.

dgottlieb:
Not 100% this actually works. I'm about to run to an interview so I just wanted to float something out there. Will double check more thoroughly when I get back.


dgottlieb:
Verified this seems to work

PR #129, opened 2025-04-28
PTZ keyboard controls example script

seanavery:
Needs to wait for PTZ do-command interface to get merged

Useful to mimic how front-end can use do command API and moving the camera around to make sure we are getting new frames when manual testing.


nicksanford (approved):
This is awesome. PTZ has been added back.

nicksanford:
> This is awesome. PTZ has been added back.

If you want to make one or two more improvements, you could add zoom and also print the controls to the user on startup.

Not needed for merge, though.

examples/ptz-ctrls/client.go
```diff
@@ -0,0 +1,174 @@
...
+	rtspCam1, err := camera.FromRobot(machine, "rtsp-cam-1")
+	if err != nil {
+		logger.Error(err)
+		return
+	}
+	rtspCam1ReturnValue, err := rtspCam1.Properties(context.Background())
+	if err != nil {
+		logger.Error(err)
+		return
+	}
```
nicksanford, line 53: [suggestion]

1. I don't think you need to query the camera
2. Let's make this take the secure credentials and the ptz resource name as CLI args. That way users can use this example without needing to change the code.

seanavery: Removed camera query.

Added .env wrapper for secrets.

PR #132, opened 2025-04-29
[RSDK-10531] RTSP preview image fallback

seanavery:
## Description
- Adds `fetchImageFromRTSPURL` helper to the `dis` service for image look-up from RTSP URI
- Fallback logic for preview do-command
## Demo

https://github.com/user-attachments/assets/951c686c-3bee-4af5-b303-2a62908d3af0

### Automated Tests
Using `ServerHandler` from the `rtsp_test.go` to mock the RTSP fallback
- Test that rtsp fallback works in snapshot failure cases ✅ 
- Test error case where snapshot and rtsp fail ✅ 

### Manual Tets
- Successful snapshot URI image lookup ✅ 
- Successful rtsp URI image look up ✅ 
- Concurrent RTSP preview lookups with multiple Control tab sessions ✅


viamonvif/service.go
```diff
@@ -175,21 +177,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+			dataURL, err := downloadPreviewImage(ctx, dis.logger, snapshotURI)
+			if err == nil {
+				dis.logger.Debugf("Successfully fetched image from snapshot URI: %s", snapshotURI)
+				return map[string]interface{}{
+					"preview": dataURL,
+				}, nil
+			}
+			dis.logger.Warnf("failed to fetch image from snapshot URI: %s, error: %v", snapshotURI, err)
+			snapshotErr = fmt.Errorf("snapshot error: %w", err)
+		} else {
```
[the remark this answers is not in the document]
seanavery: Nvm we do not want to early return -- keep as is.

viamonvif/service.go
```diff
@@ -287,6 +311,55 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+			// Attempt to get an image from the RTSP camera
+			img, metadata, err := camera.Image(ctx, "image/jpeg", nil)
+			if err == nil {
+				logger.Infof("Received image with metadata: %v", metadata)
+				dataURL := formatDataURL(metadata.MimeType, img)
+				return dataURL, nil
+			}
+			logger.Debugf("Failed to get image from RTSP camera: %v", err)
+		case <-timeoutChan:
+			return "", errors.New("timeout while trying to get image from RTSP camera")
```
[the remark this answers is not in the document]
seanavery: Also include last seen RTSP err

test_utils.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/bluenviron/gortsplib/v4/pkg/description"
+	"github.com/bluenviron/gortsplib/v4/pkg/format"
+	"github.com/bluenviron/mediacommon/pkg/codecs/h264"
+	"go.viam.com/rdk/logging"
+	"go.viam.com/test"
+	"go.viam.com/utils"
+)
+
+// NewH264ServerHandler creates a new H264 server handler for testing.
+func NewH264ServerHandler(
```
seanavery, line 22: Created `test_utils.go` so that we could export fake RTSP helpers for use in the viamonvif package. (Cannot export across packages with `_test` files)

viamonvif/service.go
```diff
@@ -287,6 +312,53 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+	}()
+
+	retryInterval := rtspImageInterval
+	timeout := rtspPollTimeout
+	ticker := time.NewTicker(retryInterval)
+	defer ticker.Stop()
+	timeoutChan := time.After(timeout)
+	var imageErr error
+	for {
+		select {
```
hexbabe, line 345: is this `select` missing a `ctx.Done()` case?

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+				"preview": dataURL,
+			}, nil
 		}
-		return map[string]interface{}{
-			"preview": dataURL,
-		}, nil
+		dis.logger.Warnf("failed to fetch image via RTSP for URL: %s, error: %v", previewReq.rtspURL, err)
+		rtspErr = fmt.Errorf("RTSP error: %w", err)
+
+		return nil, fmt.Errorf("both snapshot and RTSP fetch failed: %w; %w", snapshotErr, rtspErr)
```
hexbabe, line 217: is `errors.Join()` better here than two %w formatters

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+				dis.logger.Debugf("Successfully fetched image from snapshot URI: %s", snapshotURI)
+				return map[string]interface{}{
+					"preview": dataURL,
+				}, nil
+			}
+			dis.logger.Warnf("failed to fetch image from snapshot URI: %s, error: %v", snapshotURI, err)
+			snapshotErr = fmt.Errorf("snapshot error: %w", err)
+		} else {
+			dis.logger.Warnf("snapshot URI not found for RTSP URL: %s", previewReq.rtspURL)
+			snapshotErr = errors.New("snapshot URI not found")
```
hexbabe, line 202: should we add the url to the error itself too

test_utils.go
nicksanford, line 22:
```suggestion
func NewMockH264ServerHandler(
```

nicksanford: so it is clear this is for testing only

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+			if err == nil {
+				dis.logger.Debugf("Successfully fetched image from snapshot URI: %s", snapshotURI)
+				return map[string]interface{}{
+					"preview": dataURL,
+				}, nil
+			}
+			dis.logger.Warnf("failed to fetch image from snapshot URI: %s, error: %v", snapshotURI, err)
+			snapshotErr = fmt.Errorf("snapshot error: %w", err)
+		} else {
+			dis.logger.Warnf("snapshot URI not found for RTSP URL: %s", previewReq.rtspURL)
```
nicksanford, line 201: If we succeed in deriving the image from rtsp we don't need to warn here. Lets downgrade this to debug

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+		if found {
+			dis.logger.Debugf("attempting to fetch image from snapshot URI: %s", snapshotURI)
+			dataURL, err := downloadPreviewImage(ctx, dis.logger, snapshotURI)
+			if err == nil {
+				dis.logger.Debugf("Successfully fetched image from snapshot URI: %s", snapshotURI)
+				return map[string]interface{}{
+					"preview": dataURL,
+				}, nil
+			}
+			dis.logger.Warnf("failed to fetch image from snapshot URI: %s, error: %v", snapshotURI, err)
```
nicksanford, line 198: this might still succeed, lets downgrade to debug

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+
+		var snapshotErr, rtspErr error
+
+		// Attempt to fetch the image from the snapshot URI
 		dis.rtspToSnapshotURIsMu.Lock()
 		snapshotURI, found := dis.rtspToSnapshotURIs[previewReq.rtspURL]
 		dis.rtspToSnapshotURIsMu.Unlock()
-		if !found {
-			return nil, fmt.Errorf("snapshot URI not found for %s", previewReq.rtspURL)
+
```
nicksanford, line 188: this is more ideomatic in go.

viamonvif/service.go
```diff
@@ -175,21 +178,43 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
 		if err != nil {
 			return nil, err
 		}
+
```
nicksanford, line 181: Can we please refactor this into 1 method on `dis` with this signature:
```go
func (dis *rtspDiscovery) preview(ctx) (string, error) {
}
```
where `string` is `dataURL` 
And that method handles looking up the `snapshotURI`  and falling back to `fetchImageFromRTSPURL` if necessary?

That way this code becomes:
```
dataURL, err := dis.preview(ctx)
if err != nil {
return nil, err
}

return map[string]interface{}{"preview": dataURL}, nil
```

It also makes it much easier to move the preview functionality into a cli command if we want to.

viamonvif/service.go
```diff
@@ -287,6 +312,53 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
 }
 
+// fetchImageFromRTSPURL fetches the image from the rtsp URL and returns it as a data URL.
+func fetchImageFromRTSPURL(ctx context.Context, logger logging.Logger, rtspURL string) (string, error) {
+	// Wrap viamrtsp.Config in a resource.Config
+	rtspConfig := viamrtsp.Config{
+		Address: rtspURL,
+	}
+	resourceConfig := resource.Config{
+		Name:                "camera_name", // Replace with the actual camera name
```
nicksanford, line 322: do we need to actually replace this with the camera name or does it not matter?

viamonvif/service.go
```diff
@@ -287,6 +311,55 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+	}()
+
+	retryInterval := rtspImageInterval
+	timeout := rtspPollTimeout
+	ticker := time.NewTicker(retryInterval)
+	defer ticker.Stop()
+	timeoutChan := time.After(timeout)
+
+	for {
+		select {
```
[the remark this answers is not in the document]
nicksanford: We need to check before we go into the select.
```suggestion
	for {
	if ctx.Err() != nil {
	return "", ctx.Err()
	}
		select {
```

viamonvif/service.go
nicksanford: yes it is :)

viamonvif/service.go
```diff
@@ -287,6 +312,53 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+	defer ticker.Stop()
+	timeoutChan := time.After(timeout)
+	var imageErr error
+	for {
+		select {
+		case <-ticker.C:
+			// Attempt to get an image from the RTSP camera
+			img, metadata, err := camera.Image(ctx, imageReqMimeType, nil)
+			if err == nil {
+				logger.Infof("Received image with metadata: %v", metadata)
```
nicksanford, line 350: This can be debug... if we are failing a few times, the user doesn't need to take action on this

viamonvif/service_test.go
```diff
@@ -213,6 +215,69 @@ func TestDoCommandPreview(t *testing.T) {
...
+
+		command := map[string]interface{}{
+			"command": "preview",
+			"attributes": map[string]interface{}{
+				"rtsp_address": rtspAddr,
+			},
+		}
+		result, err := dis.DoCommand(ctx, command)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, len(result["preview"].(string)), test.ShouldBeGreaterThan, 0)
```
nicksanford, line 256: Shouldn't you be able to parse the image and have it be deterministic? Not just greater than 0?

viamonvif/service_test.go
```diff
@@ -213,6 +215,69 @@ func TestDoCommandPreview(t *testing.T) {
...
+			PacketizationMode: 1,
+			SPS: []uint8{
+				0x67, 0x64, 0x00, 0x15, 0xac, 0xb2, 0x03, 0xc1,
+				0x1f, 0xd6, 0x02, 0xdc, 0x08, 0x08, 0x16, 0x94,
+				0x00, 0x00, 0x03, 0x00, 0x04, 0x00, 0x00, 0x03,
+				0x00, 0xf0, 0x3c, 0x58, 0xb9, 0x20,
+			},
+			PPS: []uint8{0x68, 0xeb, 0xc3, 0xcb, 0x22, 0xc0},
+		}
+		h, closeFunc := viamrtsp.NewH264ServerHandler(t, forma, bURL, logger)
```
nicksanford, line 235: if this test fails it won't close

```suggestion
		h, closeFunc := viamrtsp.NewH264ServerHandler(t, forma, bURL, logger)
		defer closeFunc()
```

viamonvif/service.go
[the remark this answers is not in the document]
seanavery: Added `ctx.Err` check

viamonvif/service.go
seanavery: Yes thanks, added `ctx.Done` case

viamonvif/service.go
seanavery: Like that, changed.

viamonvif/service.go
seanavery: Added

test_utils.go
seanavery: Good call, changed.

viamonvif/service.go
seanavery: should i use `ok` instead of `found`?

viamonvif/service.go
seanavery: Good call, using `preview` helper now.

Everything was stuffed in the DoCommand because we only have one do-command handler.

viamonvif/service.go
seanavery: I am no longer concerned about the actual name used here matching up with discovery.

Also, since we are not using a robot harness, I am not worried about name collisions.

seanavery: Using stock name of `tmp-camera`

viamonvif/service.go
seanavery: Thanks

viamonvif/service_test.go
seanavery: Yes thank you

viamonvif/service_test.go
seanavery: True, checking size now.

viamonvif/service.go
```diff
@@ -175,27 +178,56 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+// preview fetches a camera image from the given RTSP URL.
+func (dis *rtspDiscovery) preview(ctx context.Context, rtspURL string) (string, error) {
+	var snapshotErr, rtspErr error
+
+	// Attempt to fetch the image from the snapshot URI
+	dis.rtspToSnapshotURIsMu.Lock()
+	snapshotURI, ok := dis.rtspToSnapshotURIs[rtspURL]
+	dis.rtspToSnapshotURIsMu.Unlock()
+
+	if ok {
```
seanavery, line 204: Set `snapshotErr` and debug log if `snapshotURI` not found

viamonvif/service.go
```diff
@@ -175,27 +178,56 @@ func (dis *rtspDiscovery) DoCommand(ctx context.Context, command map[string]inte
...
+		dis.logger.Debugf("attempting to fetch image from snapshot URI: %s", snapshotURI)
+		dataURL, err := downloadPreviewImage(ctx, dis.logger, snapshotURI)
+		if err == nil {
+			dis.logger.Debugf("Successfully fetched image from snapshot URI: %s", snapshotURI)
```
seanavery, line 208: [nit] lowercase

viamonvif/service_test.go
```diff
@@ -169,7 +171,7 @@ func TestDoCommandPreview(t *testing.T) {
 		test.That(t, result["preview"], test.ShouldEqual, "data:image/jpeg;base64,bW9ja0ltYWdlRGF0YQ==")
 	})
 
-	t.Run("Test preview command with invalid RTSP URL", func(t *testing.T) {
+	t.Run("Test preview command snapshot download with missing cache mapping", func(t *testing.T) {
```
seanavery, line 174: Improve description here

viamonvif/service.go
seanavery: Already have this handled.

viamonvif/service.go
```diff
@@ -287,6 +319,58 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+	timeout := rtspPollTimeout
+	ticker := time.NewTicker(retryInterval)
+	defer ticker.Stop()
+	timeoutChan := time.After(timeout)
+	var imageErr error
+	if err := ctx.Err(); err != nil {
+		return "", fmt.Errorf("context already canceled before fetching image: %w", err)
+	}
+	for {
+		select {
```
nicksanford, line 355: This check before going into the select statement is needed in order to ensure that we always return early return if the context is cancelled.

Otherwise if both  `ticker.C` fires and `ctx` is cancelled at the same time, `select` could take the `ticker.C` path.


```suggestion
	for {
	        if ctx.Err() != nil {
	            return "", fmt.Errorf("context canceled while fetching image from RTSP camera: %w", ctx.Err())
	        }
		select {
```

viamonvif/service_test.go
nicksanford: where are you checking the size?

viamonvif/service_test.go
```diff
@@ -213,6 +215,69 @@ func TestDoCommandPreview(t *testing.T) {
...
+
+		command := map[string]interface{}{
+			"command": "preview",
+			"attributes": map[string]interface{}{
+				"rtsp_address": rtspAddr,
+			},
+		}
+		result, err := dis.DoCommand(ctx, command)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, len(result["preview"].(string)), test.ShouldEqual, 7435)
```
nicksanford, line 257: can you please decode the image and assert on it's bounds so we can be confident that we actually implemented a preview?

viamonvif/service.go
```diff
@@ -287,6 +319,58 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
...
+	}()
+
+	retryInterval := rtspImageInterval
+	timeout := rtspPollTimeout
+	ticker := time.NewTicker(retryInterval)
+	defer ticker.Stop()
+	timeoutChan := time.After(timeout)
+	var imageErr error
+	if err := ctx.Err(); err != nil {
+		return "", fmt.Errorf("context already canceled before fetching image: %w", err)
```
nicksanford, line 352: This check is not necessary. Just add the check (with an appropriate error) inside the loop before the select.

viamonvif/service.go
nicksanford: actually , nevermind. It shouldn't matter in this case.

viamonvif/service.go
seanavery: Ok, yes removed extra check. I think the ` <-ctx.Done()` case should handle things.

nicksanford (approved):

PR #134, opened 2025-05-05
Add .env wrapper to discovery example script

seanavery:
## Description

Just added env var wrapper so devs do not need to change code to get things running.

## Tests
- Tested without .env and using environment variable ✅ 
- Tested error messages when env vars missing ✅ 
- Tested without onvif username and password ✅ 
- Tested with onvif username password ✅


hexbabe (approved):
LGTM

PR #140, opened 2025-05-27
[RSDK-10781] - Validate snapshot return mimetype

seanavery:
## Description

[RSDK-10781](https://viam.atlassian.net/browse/RSDK-10781)
We encountered an issue with the preview do-command returning the log-in webpage instead of the image, even with valid credentials. This PR simply validates the mimetype of the response to make sure we are receiving a valid image.

## Testing

Added test case where snapshot endpoint returns html.

## Refs

[parseMediaType](https://pkg.go.dev/mime#ParseMediaType)

[RSDK-10781]: https://viam.atlassian.net/browse/RSDK-10781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/viamonvif/service.go
+++ b/viamonvif/service.go
@@ -309,8 +309,13 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
 	}
 
 	contentType := resp.Header.Get("Content-Type")
-	if !isImageMIMEType(contentType) {
-		return "", fmt.Errorf("snapshot URI returned non-image content type: %s", contentType)
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse content type %s: %w", contentType, err)
+	}
+	isImage := strings.HasPrefix(mediaType, "image/")
+	if !isImage {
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -139,7 +139,7 @@ tool-install:
 gofmt:
 	gofmt -w -s .
 
-lint: gofmt tool-install build-ffmpeg update-rdk
+lint: gofmt tool-install build-ffmpeg
```
hexbabe, line 142: intentional? would upgrading block this PR?

viamonvif/service.go
```diff
@@ -319,6 +323,14 @@ func downloadPreviewImage(ctx context.Context, logger logging.Logger, snapshotUR
 	return dataURL, nil
 }
 
+func isImageMIMEType(mimeType string) bool {
```
hexbabe, line 326: Do we expect to use this helper elsewhere in the future? If no, maybe we should just keep the logic in-line

seanavery: Sounds good to me, makes it easier to read without jumping around.

PR #141, opened 2025-05-29
RSDK-10807 - Handle concurrent RTSP preview fallback attempts

seanavery:
## Description

Currently, we hardcode `tmp-camera` as the RTSP preview look up resource name. This causes issue with concurrent spin ups. This PR simply ensures uniqueness for each resource name so we can spin up arbitrary number of concurrent resources.

`camera_name`-`timestamp` -> `tmp-camera-1684263518248967000`

## Tests

Added concurrency test case to the service to make sure the issue has been fixed.

### DIFF
--- a/viamonvif/service.go
+++ b/viamonvif/service.go
@@ -3,10 +3,8 @@ package viamonvif
 
 import (
 	"context"
-	"crypto/rand"
 	"crypto/tls"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -337,10 +335,7 @@ func fetchImageFromRTSPURL(ctx context.Context, logger logging.Logger, rtspURL s
 	rtspConfig := viamrtsp.Config{
 		Address: rtspURL,
 	}
-	uniqueName, err := generateUniqueName("tmp-camera")
-	if err != nil {
-		return "", fmt.Errorf("failed to generate unique camera name: %w", err)
-	}
+	uniqueName := generateUniqueName("tmp-camera")
 	resourceConfig := resource.Config{
 		Name:                uniqueName,
 		API:                 camera.API,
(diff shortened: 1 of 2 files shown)


viamonvif/service.go
```diff
@@ -3,8 +3,10 @@ package viamonvif
 
 import (
 	"context"
+	"crypto/rand"
```
hexbabe, line 6: RAND!!!

viamonvif/service.go
```diff
@@ -377,6 +384,18 @@ func fetchImageFromRTSPURL(ctx context.Context, logger logging.Logger, rtspURL s
 	}
 }
 
+// generateUniqueName creates a unique name by adding timestamp and random bytes
+func generateUniqueName(prefix string) (string, error) {
```
hexbabe, line 388: This is a nice helper. Could we write unit tests for it?

seanavery: Good call, added unit test

viamonvif/service.go
```diff
@@ -377,6 +384,18 @@ func fetchImageFromRTSPURL(ctx context.Context, logger logging.Logger, rtspURL s
...
+	randomBytes := make([]byte, rtspNameSaltLength)
+	if _, err := rand.Read(randomBytes); err != nil {
+		return "", err
+	}
+	timestamp := time.Now().UnixNano()
+	randomHex := hex.EncodeToString(randomBytes)
+	uniqueName := fmt.Sprintf("%s-%d-%s", prefix, timestamp, randomHex)
```
nicksanford, line 395: I don't think we are going to get 64 calls in the same nanosecond. If we need to make the names unique for practical purposes, the nanosecond timestamp should be sufficient.

That also eliminates the need for this function to return an error.

viamonvif/service_test.go
```diff
@@ -327,6 +329,93 @@ func TestDoCommandPreview(t *testing.T) {
 		test.That(t, err.Error(), test.ShouldContainSubstring, "snapshot URI returned non-image")
 		test.That(t, err.Error(), test.ShouldContainSubstring, "text/html")
 	})
+
+	t.Run("Test concurrent preview commands with RTSP fallback", func(t *testing.T) {
```
nicksanford, line 333: why would this test fail before?

nicksanford (changes requested):

nicksanford:
I don't understand why creating multiple cameras with the same name would cause a change in behavior.

Can you explain why this is necessary?

nicksanford (approved):

PR #148, opened 2025-07-30
[RSDK-11506] Fix readme link for video-store

hexbabe:


randhid (approved):

randhid:
Remember to run `viam module update` - nowadays it should also run when you make a new release - but I don't think this is worth a release on its own, so we have that function to update the config card text.

hexbabe:
> Remember to run `viam module update` - nowadays it should also run when you make a new release - but I don't think this is worth a release on its own, so we have that function to update the config card text.

Screenshot 2025-07-30 at 5 16 21 PM
ran `viam module update` and verified that the correct section shows up in UI

PR #205, opened 2026-03-30
Reject loopback IPs in mDNS registration

seanavery:
mDNS hostnames (e.g. `ND012501000843.local`) resolving to `127.0.0.1` instead of real camera IP, causing ONVIF/PTZ endpoints to fail with `connection refused`. Root cause is in zeroconf `appendAddrs()` injecting localhost on loopback interface queries.

Root cause fix: https://github.com/viamrobotics/zeroconf/pull/7

This PR adds defense-in-depth in viamrtsp:
- `tryMDNS()`: skip registration if `deviceIP` is loopback
- `mdnsServer.Add()`: reject loopback IPs with error


SebastianMunozP (approved):
LGTM

PR #209, opened 2026-04-08
Force rtp marker bit on new timestamp

seanavery:
## Description

Some cameras (e.g. Garmin with GStreamer RTSP server) never set the RTP marker bit on H264/H265 packets. gortsplib's decoder uses the marker bit to detect access unit (frame) boundaries; without it, the NALU buffer grows indefinitely until hitting `MaxNALUsPerAccessUnit` and every frame is dropped.

Fix: buffer one RTP packet and compare timestamps. When the timestamp changes, force `Marker=true` on the previous packet before processing. Applied to both H264 and H265 paths.

## Tests

- [x] Manual test on Garmin camera (no more NALU overflow errors)
- [x] Verify normal cameras with proper marker bits still work (marker already set, forced `true` is a no-op)

## Review

1 approval

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -585,15 +585,9 @@ func (rc *rtspCamera) initH264(session *description.Session) (err error) {
 	}
 
 	params := [][]byte{f.SPS, f.PPS}
-	onPacketRTP := func(pkt *rtp.Packet) {
-		pts, ok := rc.client.PacketPTS2(media, pkt)
-		if !ok {
-			rc.logger.Debug("no pts found for packet")
-			return
-		}
-		if publishToWebRTC != nil {
-			publishToWebRTC(pkt, pts)
-		}
+
+	// processPacket handles decoding and storing a single RTP packet.
+	processPacket := func(pkt *rtp.Packet, pts int64) {
 		au, err := rtpDec.Decode(pkt)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


rtsp.go
```diff
@@ -1485,3 +1483,20 @@ func (rc *rtspCamera) NextPointCloud(_ context.Context, _ map[string]interface{}
...
+
+// wrapWithMarkerFromTimestamp wraps an RTP packet handler to force the marker bit
+// when the RTP timestamp changes. Some cameras (e.g. GStreamer-based) never set
+// the marker bit, causing gortsplib to buffer NALUs indefinitely. Timestamp
+// changes indicate a new access unit, so we mark the previous packet as complete.
+func wrapWithMarkerFromTimestamp(handler func(pkt *rtp.Packet)) func(pkt *rtp.Packet) {
+	var pendingPkt *rtp.Packet
+	return func(pkt *rtp.Packet) {
+		if pendingPkt != nil {
+			if pkt.Timestamp != pendingPkt.Timestamp {
```
SebastianMunozP, line 1495: Should this boolean operator be > instead of !=?

seanavery: Yeah probably a good idea to move to `>`, changing

seanavery: Actually, RTP timestamps are uint32 and wrap around 0xFFFFFFFF to 0x00000001 so probably want to leave as is

https://www.rfc-editor.org/rfc/rfc3550#section-5.1

SebastianMunozP (approved):
LGTM

PR #211, opened 2026-04-30
Whitelist benchmark IP range

seanavery:
## Description

Quick patch to whitelist the benchmark IP address range. This range is used by Raymarine MFD and cameras.

[RFC 2544](https://www.ietf.org/rfc/rfc2544.txt)

[Raymarine Camera Docs](https://www.raymarine.com/en-us/download/cam300-manuals)

## Testing

Verified patch works on customer machine with live Raymarine cam setup.

viamrtsp `0.5.9-rc1`


ale7714 (approved):

ale7714:
I think this is very reasonable overall. I think we can hold making further improvements until we have another use case to whitelist ip ranges and then, we should consider making it configurable.

johnwalicki (approved):
Always a win when we can link to the RFC, become more conformant and expand our discoverability of marine cameras.  LGTM

PR #214, opened 2026-05-19
Fix deadlock between close and reconnect

seanavery:
## Description

[Flaky test](https://github.com/viam-modules/viamrtsp/actions/runs/26051050491/job/76590738812) surfaced deadlock between reconnect and Close.

### DIFF
--- a/rtsp.go
+++ b/rtsp.go
@@ -265,9 +265,7 @@ func (rc *rtspCamera) Close(_ context.Context) error {
 	}
 	rc.cancelFunc()
 	rc.unsubscribeAll()
-	// Wait for the reconnect worker to exit before taking closeMu — the worker
-	// acquires closeMu inside reconnectClient, so holding it across Wait() would
-	// deadlock with an in-flight reconnect attempt.
+	// Wait for the reconnect worker to exit before taking closeMu
 	rc.activeBackgroundWorkers.Wait()
 	rc.closeMu.Lock()
 	defer rc.closeMu.Unlock()
--- a/rtsp.go
+++ b/rtsp.go
@@ -208,12 +208,16 @@ type rtspCamera struct {
 	lazyDecode       bool
 	iframeOnlyDecode bool
 
+	// closeMu guards client/h264Media/rawDecoder and lifecycle of mimeHandler/avFramePool.
+	// Writers Lock during teardown/reconnect, readers RLock for the duration of their work.
+	// Outermost lock: take before auMu/latestFrameMu/subsMu.
 	closeMu      sync.RWMutex
 	videoRequest *videoRequest
-	auMu         sync.Mutex
-	au           [][]byte
-	client       *gortsplib.Client
-	rawDecoder   *decoder
+	// auMu guards rc.au (lazy AU buffer).
+	auMu       sync.Mutex
+	au         [][]byte
+	client     *gortsplib.Client
+	rawDecoder *decoder
 	// h264Media is the RTSP media track for H264.
 	h264Media *description.Media
 	// firSeqNum holds the last FIR sequence number (0–255), wraps per RFC 5104.
@@ -224,8 +228,7 @@ type rtspCamera struct {
 
 	activeBackgroundWorkers sync.WaitGroup
 
-	// latestFrameMu protects critical sections where frame state changes (e.g. ref counting) need to be atomic
-	// with swapping out the latest frame.
+	// latestFrameMu guards latestFrame ref counting and latestFrameCache.
 	latestFrameMu    sync.Mutex
 	latestMJPEGBytes atomic.Pointer[[]byte]
 	latestFrame      *avFrameWrapper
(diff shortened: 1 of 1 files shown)


rtsp.go
```diff
@@ -264,9 +264,11 @@ func (rc *rtspCamera) Close(_ context.Context) error {
 		rc.logger.Errorf("error removing camera from global registry: %s", err.Error())
 	}
 	rc.cancelFunc()
-	rc.closeMu.Lock()
 	rc.unsubscribeAll()
+	// Wait for the reconnect worker to exit before taking closeMu
 	rc.activeBackgroundWorkers.Wait()
+	rc.closeMu.Lock()
```
dgottlieb, line 280: @benjirewis can you and @seanavery discuss this locking w.r.t active background workers and expectations around in flight requests during close?

benjirewis: I haven't looked through this whole method/module, but I'd "normally" expect to see:
```
rc.closeMu.Lock()
rc.cancelFunc()
rc.closeMu.Unlock()
```
As a pattern for "request that all ongoing work stop and then wait for that ongoing work to actually stop." Let me know if you want to sync offline @seanavery. You'd have to walk me through what `Close` is trying to do here, but what you have currently _looks_ a bit off.

seanavery: hey @benjirewis @dgottlieb, walking through my understanding of what was happening here

In the original Close:
```go
rc.cancelFunc()
rc.closeMu.Lock()
rc.unsubscribeAll()
```

`clientReconnectBackgroundWorker` takes `closeMu` inside `reconnectClient` for each transport attempt. So it would block on the lock that `Close` was holding, while `Close` was blocked on `Wait()` waiting for that same worker to exit.

The test sets `reconnectIntervalDuration = 10ms`, which is why this surfaced so reliably. (Prod uses 5s)

The suggested `Lock, cancel, Unlock, Wait` pattern works nicely, the key thing is not holding the lock over `Wait()`. I didn't wrap `cancelFunc()` itself because `context.CancelFunc` is goroutine safe and the RWMutex around `closeConnection()` (the bit that does `client.Close() rc.client = nil` etc) already serializes teardown vs concurrent readers. Happy to switch to the canonical ordering for style, lmk.

benjirewis: Sean and I spoke offline a bit about better usage of `closeMu` and `cancelCtx` + better documentation around the four mutexes on the `rtspCamera` struct.

seanavery: Thanks @benjirewis. First pass at adopting your suggested pattern, and it seems to have turned out nicely

Bonus: holding closeMu over cancel let me wrap SubscribeRTP in RLock, which closed out a separate orphan subscription race and let me drop the mu dance around the FIR call.

rtsp.go
```diff
@@ -208,12 +208,14 @@ type rtspCamera struct {
 	lazyDecode       bool
 	iframeOnlyDecode bool
 
+	// closeMu guards gortsplib client, h264Media, rawDecoder and lifecycle of mimeHandler/avFramePool.
```
benjirewis, line 211: [nit] I'd put the fields a mutex guards underneath the mutex in the struct definition. Like:

```
fooMu sync.Mutex
foo *foo
fooClient *Client
fooValue *Value

barMu sync.Mutex
bar *bar
barClient *Client
barValue *Value
```

rtsp.go
benjirewis: I'm surprised you're calling `closeMu.Lock()` twice in this function. Does everything below this line need `closeMu` held? If so, is there reason not to include it in the critical section above instead of this second one?

rtsp.go
```diff
@@ -1074,13 +1095,8 @@ func (rc *rtspCamera) SubscribeRTP(
 	// Send an RTCP FIR to request an IDR keyframe from the camera so the new subscriber
 	// receives a decodable frame immediately. This is best-effort: if the camera ignores FIR
 	// or the client is mid-reconnect, the subscriber will receive an IDR naturally.
-	rc.closeMu.RLock()
```
benjirewis, line 1077: You're removing these lines because we now hold the `RLock` over the whole method? Seems fine to me just checking.

rtsp.go
seanavery: Good call. Punted on closeMu initially since its fields were scattered all over the struct, but you're right that clustering them reads way better

rtsp.go
seanavery: Yeah, not entirely happy with this, but it seems to be necessary for now:
- first lock makes cancelFunc atomic w the RLock holders
- cannot hold lock over Wait (reconnect loop deadlock)
- second lock is for general teardown

Lmk if a comment near the second lock would help

I think it is possible to refactor reconnect to not need the closeMu but seems out of scope for this PR.

benjirewis (commented):
The usages of the `closeMu` certainly look more "normal" to me now, but I won't claim to have all the context to approve here. I'm still suspicious of the need to lock twice in `Close`, but your added documentation and the more consistent pattern in using/setting the client object make sense to me!

